PR #8, opened 2020-04-06
[WIP] Avena ansible

abalmos:
This is a work in progress tracking PR for Avena's initial Ansible scripts.

I do believe that (most?) everything we need is here now, but these scripts have issues still. I will fix those, but leaving this here for others to start getting some exposure.

This does not call `docker-compose` -- we need think about how to appropriately structure that.
Ideas:
- have nodes pull a git repo with the docker-compose.yml file in it and run `docker-compose up -d`. Then changes to the docker services would not be through Ansible, but rather by periodically pulling a git.

- have Ansible copy a docker-compose.yml file from somewhere and start it. We would have a playbook that would update those docker-compose files.

Something to keep in mind: We will likely have a large fleet of "avena" nodes, but there could be many configurations of the docker-compose services ... thinking ISOBlue HD, wifi test, etc.


abalmos:
Merging this without review ... it's our starting point :shrug:

PR #24, opened 2020-04-21
WIP OADA Integration

aaron97neu:
WIP PR tracking the integration of a db to track gps points and a corresponding node.js script to push them to an OADA instance

@abalmos I would like you to specifically inspect `oada_upload.js/ts`, looking for issues in both the desired program flow defined in the comments around line 100 and the actual program flow. There are issues with how I am using and chaining promises, how can i do that better?

Todos in no particular priority:
- Integrate waiting for internet connection feature **Partially done in typescript migration**
- ~Integrate batching of multiple points into one `PUT` request~ **Done**
- Investigate high CPU usage (Busy waiting when sleeping?) and fix
- Improvement in overall fault tolerance and error checking **Partially done in typescript migration**
- Testing when integrated with @wang701 's features
- Generalize so that we can upload CAN data and heartbeat messages (or any other generic data we decide to collect). Will require changes to DB schema
- Increase security (Better way to store OADA tokens and db passwords than in `dockerfile` and `.env`?)
- ~Make STDOUT reasonable but still informative. 10+ lines per data point is way too much and may be contributing to high CPU usage~ **Done**
- ~Merge `.env` and `oada_enviroment.env` files~ **Done**

Possible additional todos:
- ~Switch to timescaledb or influxdb~ **Done**
- ~Switch to yarn instead of npm~ **Done**


facastiblancor:
Looks good to me. Is the db password still an issue after migrating to timescale? Would it be possible to use strong passwords or use another authentication method?

aaron97neu:
> Looks good to me. Is the db password still an issue after migrating to timescale? Would it be possible to use strong passwords or use another authentication method?

@facastiblancor Since timescale is just an extension over postgres I do not expect the special-char-in-password problem to be solved. I suspect it has something to do with the `node-pg` yarn package I am using which has not changed. Andrew and I talked about the use of a weak or even nonexistent password previously. The conclusion we came to (@abalmos correct me if I am wrong) was that since the ISOBlue is running inside the VPN no one should have access to the network (and therefore the database) unless the VPN in compromised. If VPN is compromised that means that our vault that controls access to the VPN has been compromised and we are really in big trouble.

That being said avoiding a single point of failure is important and it still makes me uncomfortable. I have a few features that I still want to add to this including using docker secrets as well as investigating if I can restrict access to the database by non docker containers

abalmos:
I'm not entirely convinced a password is overly helpful here. If you can access postgres over the loopback then you own ISOBlue and any secrets stored in its configuration files.

Now, if we expose timescale to the wireguard interface (which could be quite interesting), then we most definitely should have a password on that connection.

I assume postgres shares mysql's auth model: it's a user+password+hostname combo?

I suggest we move this to a dedicated issue. We will need to figure out what in the project will be responsible for bringing up postgres/timescale. It will almost certainly not be limited to GPS. I also think we are going to need to explore something like docker's wait-for-it.sh fix, so containers wait for their needed services.

software/oada_upload/src/oada_upload.ts
```diff
@@ -0,0 +1,387 @@
...
+        },
+        location: {
+          lat: p.lat,
+          lng: p.lng,
+        },
+      };
+      
+    });
+    // pEachSeries executes the following code for each bucket (path) that we created sequentially 
+    await pEachSeries(Object.keys(data), async (path) => {
```
abalmos, line 309: You can likely drop `pEachSeries` and replace with a `for (let path in data) { ...` because `oada-client` will limit request concurrency. Not sure its a required change though.

software/oada_upload/src/oada_upload.ts
```diff
@@ -0,0 +1,387 @@
...
+      var res;
+
+      // Attempt the put to OADA
+      try {
+        //console.debug(`GPS ID ${data[path].epochs.join(',')} to OADA ${path}`);
+        console.debug(`${data[path].epochs.length} points to OADA ${path}`);
+        res = await oada.put({
+          tree: isoblueDataTree,
+          path,
+          data: JSON.parse(JSON.stringify({data: data[path].locations})),
```
abalmos, line 320: I think `JSON.parse(JSON.stringify(thing))` is a no-op unless:

1. You expect `thing` to change between the call of `oda.put` and it completin
2. `thing` has no JSON serializable things in it that is breaking `@oada/client` (if so, please report the bug to @oada/client)

Otherwise, I think just:

```suggestion
          data: {data: data[path].locations},
```

should work

software/oada_upload/src/oada_upload.ts
```diff
@@ -0,0 +1,387 @@
...
+//   - Use p-queue to control the calls to OADA / DB (any Promise), and then use
+//   queue length to make decisions on weather or not to continue searching.
+//   p-queue comes with the ability to have a call back when the queue gets
+//   below a certain size / empty. You could fetch from the DB in that case, and
+//   the load up OADA calls.
+function sleep(ms: number): Promise<void> {
+  return new Promise((resolve) => {
+    setTimeout(resolve, ms);
+  });
+}
```
abalmos, line 382:
```suggestion

```

software/oada_upload/src/oada_upload.ts
```diff
@@ -0,0 +1,387 @@
...
+        } catch (e) {
+          console.error(`FATAL: Error updating database with sent information: `, (<Error>e).message);
+          // Do something with the error?
+          // This seems pretty fatal, exit and hope that when docker restarting the process fixes it
+          process.exit(-1);
+        }
+      }
+      // Wait half a second between uploads to not overwhelm the isoblue, OADA server, or the programmer trying to debug stuff
+      // Should be removed enentually. Likely once OADA-Client has concurrency control
+      await sleep(500);
```
abalmos, line 354: I think it is safe to delete now

```suggestion
```

software/oada_upload/src/oada_upload.ts
```diff
@@ -0,0 +1,387 @@
...
+        //console.debug(`oada put finished: `, res);
+        console.debug(`OADA put finished`);
+      } catch (e) {
+        // The PUT is not a success. This is either due to a misconfiguration or lack of internet.
+        // Assume that everything is configured correctly rebuild the queue for the next try.
+        console.error(`Error Uploading to OADA: %p`, e, e.message, (<Error>e).message, ` `, res);
+        update_success = false;
+      }
+      console.debug(`Done uploading to OADA`);
+      if (update_success){ // Replace the if with a 'continue' in the previous error state?
```
abalmos, line 331: This looks right, but I am a bit confused  by it. Why not put a `continue` on line 328 and just drop the conditional here? I am mostly asking to make sure I didn't miss some nuance

abalmos: Wait, I understand the comment now :) ... It's suggesting the same thing as my comment.

software/oada_upload/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
```
abalmos, line 3: Since we are targeting ts-node, we should use `ES6` here and use real modules.

software/oada_upload/tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+    "moduleResolution": "node",
+    "target": "es6",
+    "lib": ["es2019"],
+    "esModuleInterop": true,
```
abalmos, line 7: Do things break if you delete this line? Seems better to not have it on if possible.

software/postgres/schema/gps.sql
```diff
@@ -0,0 +1,11 @@
+DROP TABLE IF EXISTS "gps";
```
abalmos, line 1: What are the "best practice" around database schema management? We will want to migrate a table structure without having to drop it one day -- with mysql I have always used a third party module to version and migrate databases, but that isn't likely to work well in our "micro-service" style design.

software/oada_upload/src/oada_upload.ts
abalmos: Err... I guess you'll want something like `await Promise.all(Object.keys(data), (path) => { ... }` instead of the `for (let path in data) { ... ` so that you wait until all of the pending requests to complete before going back and fetching more data from the database.

software/oada_upload/src/oada_upload.ts
aaron97neu: I get the following error when using your suggestion (which is how the code was originally written before the update from oada-client~0.3 to 1.0.2):
```
$ tsc
src/oada_upload.ts(320,11): error TS2322: Type '{ data: { [key: string]: Pick<Location, string | number>; }; }' is not assignable to type 'Json'.
  Types of property 'data' are incompatible.
    Type '{ [key: string]: Pick<V1, string | number>; }' is not assignable to type 'string | number | boolean | Json[] | { [prop: string]: Json; } | null | undefined'.
      Type '{ [key: string]: Pick<V1, string | number>; }' is not assignable to type 'string'.
```

My limited understanding is that the `data` property does not accept the object(?) we created but will accept a JSON object (which is what our object represents). `JSON.stringify`, however, will accept our object and put it into string form which `JSON.parse` can convert into a JSON object.

Possible we could coerce/cast our object into a JSON object directly but I did not think of that at the time

software/docker-compose.yml
```diff
@@ -0,0 +1,84 @@
...
+      - 5432:5432
+    volumes:
+      - postgres:/var/lib/postgresql/data
+      - ./postgres/schema:/docker-entrypoint-initdb.d
+  oada-upload:
+    build:
+      context: ./oada_upload
+    restart: always
+    environment:
+      NODE_TLS_REJECT_UNAUTHORIZED: "0"
```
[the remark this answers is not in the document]
aaron97neu: Yes. @tomohiroarakawa 's server had an expired SSL key and I couldn't get SSL to work on my server

software/docker-compose.yml
```diff
@@ -0,0 +1,84 @@
...
+      db_port: "5432"
+    depends_on:
+      - postgres
+  adminer: # For dev ONLY
+    image: adminer
+    restart: always
+    ports:
+      - 8080:8080
+  gps:
+    image: abalmos/gps
```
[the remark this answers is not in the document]
aaron97neu: I will remove it for the PR but how do we want to handle development tools like that and the NODE_TLS_REJECT_UNAUTHORIZED environmental var in general? Would be nice to have those available to us to switch on of needed which we are still in such an early development stage

software/oada_upload/src/oada_upload.ts
```diff
@@ -0,0 +1,387 @@
...
+  // With the db client setup, initiate a connection to the database
+  try{ 
+    await db.connect();
+  }catch (e){
+    // The databse is hosted on the docker network, so there really shouldn't be any reason we
+    // cannot connect. If we cannot connect this could mean either there is a severe 
+    // misconfiguration (in which we cannot do much) or the db is still booting 
+    // (Happens occasionally when the db booting for the first time). 
+    // TODO: If the DB is still booting, exiting to restart is fine, but should 
+    // investigate if there are ways to detect this wait before connecting
```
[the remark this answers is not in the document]
aaron97neu: Was hoping to use a modified `wait-for-postgres.sh` that they gave in those docs, however this container does not has `psql` installed and `psql` typically requires an entire postgres installation from what I can tell. Will look into a `wait-for.sh` instead

software/oada_upload/src/oada_upload.ts
aaron97neu: In order to make use of the `oada.client` concurrency we need to do this. Going to try `Promise.all` and `pMap`

software/docker-compose.yml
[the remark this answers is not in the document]
abalmos: @aaron97neu I suppose we just need to figure out how we will deploy these things. If ansible is involved, then this file wouldn't be used at it all. There would be an Ansible template -- some things may require this. For example, CAN won't always be 2 busses (JD has often has 3 and trucks/cars have 1). 

I don't think it is an issue for this PR. Just was worth getting in our heads.

software/docker-compose.yml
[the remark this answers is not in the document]
abalmos: Okay. Will be move the ISOBlue app server to oats2? `oada-srvc-docker` should be getting valid SSL certs from lets-encrypt

software/oada_upload/src/oada_upload.ts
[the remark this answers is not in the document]
abalmos: I see. It is nice to have an application specific health-check ... but I'm not sure its worth needing to ship psql around and expose any cruff that it may be carrying. 

I assume postgres will accept a query after opening the port?

software/oada_upload/src/oada_upload.ts
abalmos: Promise.all and pMap should be equal if you don't care about the concurrency (which you shouldn't because @oada/client will manage it)

software/oada_upload/src/oada_upload.ts
abalmos: Bah, that damn JSON type looked cool, but it is being a pain... Maybe just a 

`data: {data: data[path].locations} as JSON`

you might have to import JSON ... hopefully @oada/client exposes it.

If not, just leave it alone. the `as` method has zero runtime overhead though.

software/docker-compose.yml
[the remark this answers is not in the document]
aaron97neu: I suppose. I will talk to Tomo about that, will allow me to keep `oada-srvc-docker` up to date without having to rely on him. Will take it out of the docker-compose

software/oada_upload/tsconfig.json
aaron97neu: Could you explain this a little more? Why we want ES6 instead of commonjs? Simply switching that line leads to all sorts of compiler errors

software/oada_upload/tsconfig.json
aaron97neu: Does not break things, did previously. Removed

software/oada_upload/src/oada_upload.ts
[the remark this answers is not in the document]
aaron97neu: I don't know for sure but it would make sense. Worst case the this process exits, restarts, and tries again

software/oada_upload/tsconfig.json
abalmos: Sorry. This is a node script. Been doing too much UI stuff lately. `commonjs` is right

software/oada_upload/src/oada_upload.ts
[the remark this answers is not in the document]
abalmos: :+1:

software/oada_upload/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "description": "Upload Avena time-series datapoints to an OADA store",
+  "main": "src/oada_upload.ts",
+  "repository": "https://github.com/oats-group/isoblue-avena",
+  "author": "Aaron Neustedter",
+  "contributors": [
+    "Andrew Balmos <abalmos@purdue.edu>"
+  ],
+  "license": "MIT",
+  "scripts": {
+    "start": "ts-node --project tsconfig.json src/oada_upload.ts",
```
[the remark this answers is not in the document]
aaron97neu: Nope, removed

software/oada_upload/src/oada_upload.ts
```diff
@@ -0,0 +1,387 @@
...
+      //console.debug(`Processing GPS time: ${p.time_tz}`);
+
+      // Extract time and use it to create path that it will be uploaded it
+      const t = moment.unix(p.time_epoch);
+      const day = t.format('YYYY-MM-DD');
+      const hour = t.format('HH');
+      const path = `/bookmarks/isoblue/device-index/${id}/location/day-index/${day}/hour-index/${hour}`;
+      // Create a UUID that can be coarsely sorted by time of creation
+      // TODO: find a way to use the timestamp to create the id instead of the current time
+      const pId = ksuid.randomSync().string;
```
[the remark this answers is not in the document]
aaron97neu: Fixed with similar idea to your suggestion

software/oada_upload/src/oada_upload.ts
aaron97neu: This is odd...
```
Type 'JSON' is not assignable to type 'Json'.
  Type 'JSON' is not assignable to type '{ [prop: string]: Json; }'.
    Index signature is missing in type 'JSON'.
```

If you have an idea let me know, otherwise will tackle it after the concurrency stuff

software/oada_upload/src/oada_upload.ts
aaron97neu: We still sleep when we are waiting for the database to fill, going to leave it in for that and debugging purposes

software/docker-compose.yml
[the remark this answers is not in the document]
aaron97neu: Talked with Tomo, ISOBlue app server will be moved to oats2 in the near future

software/docker-compose.yml
```diff
@@ -0,0 +1,84 @@
...
+    restart: always
+    environment:
+      GPSD_URL: "172.17.0.1:2947"
+      DATABASE_URL: "postgres://avena:1ubuntu9@@postgres/avena"
+  kafka:
```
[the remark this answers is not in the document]
aaron97neu: Could you explain this a little more? How are the other docker containers setup? I assume my stuff should mirror those

software/postgres/schema/gps.sql
aaron97neu: So this is a file that you wrote that I have not . I assume it is used for the initial setup of the docker container?

For schema management, I have little experience with it and what it entails/requires. After a brief look, it appears we would be doing the same as mysql and using a 3rd party module like https://github.com/sqitchers/sqitch to do some of the work. Unsure how well it will work with our overall architecture

software/oada_upload/src/oada_upload.ts
abalmos: Where are you importing the type `JSON` from? Looks like I should have said

 `data: {data: data[path].locations} as Json`

Notice the case differences.

software/postgres/schema/gps.sql
abalmos: Okay, I see. It is a great question on how we want to manage all this. The `postgres` init stuff only runs on the very first `up`, so that wont help with migrations ...

I guess we need services to be able to:

1. Create users for themselves (say a `gps` user which write access to `gps` table and a `oada` user with read access to `gps` and write access `oada` [or whatever that table to store "sent" is called])

2. Update / create tables on service startup (or somehow right before service startup)

I am working on docker-compose.yml files, container building, etc right now in #28, but I don't have any ideas for this. For now, we can just punt and get the db setup right with the postgres init, but any updates will need a solution.

software/oada_upload/tsconfig.json
aaron97neu: It appears that the moment package breaks when this is removed. I added this back in

software/oada_upload/src/oada_upload.ts
aaron97neu: Removed with addition of concurrency

software/oada_upload/src/oada_upload.ts
aaron97neu: Done. Needed a `return` instead of a `continue` because it is not a loop

PR #42, opened 2020-07-11
Regorganization of services and software folders

aaron97neu:
It seems that at some point the services/software/docker folders got all mixed up. In reality these should have been all one folder. Reorganized and merged all of the folders


services/docker-compose.yml
```diff
@@ -1,5 +1,35 @@
...
+      oada_server_url: "server_url.example.com"
+      oada_server_token: "abc"
+      db_user: "avena"
+      db_host: "postgres"
+      db_database: "avena"
+      db_password: "password"
+      db_port: "5432"
+    depends_on:
+      - postgres
   kafka:
```
aaron97neu, line 33: What is the future of these Kafka containers? Are we using them?

services/docker-compose.yml
```diff
@@ -1,5 +1,35 @@
-version: '2'
```
aaron97neu, line 1: There is a junja2 template version of this our other repo. Which should we use?

aaron97neu:
Should be a pretty easy PR, however have a few questions about the `docker-compose.yml`

services/docker-compose.yml
abalmos: I think their future is no more, I guess that's why I never merged them into services.

services/docker-compose.yml
abalmos: What docker-compose file is this? The one that gets deployed is produced by Ansible I believe?

aaron97neu:
The current state of the repository has two folders: `software` and `services`. `services` has two old folders that should have been deleted (`services/deployment` and `services/wireguard`, both related to my initial stab at ansible). The `services/docker` folder contains all of @wang701's Kafka stuff. `software` contains these exact  same files, plus my `oada_upload` stuff. 

Both `software` and `services/docker` had a `docker_compose.yml` so I merged them

@abalmos long answer to a short question I guess is that it's just a reorg deleting redundant copies of the Kafka stuff

abalmos:
I see. That's interesting (sorry I didn't look closer before). Maybe a bad merge brought `software` back? I renamed `software` to `services` in https://github.com/OATS-Group/isoblue-avena/commit/1bc068b470fbadcb54c101d7f31bec745d0af7b9

I am in favor of deleting `software`, `services/deployment`, `services/wireguard`, and `services/docker`. I also think all `docker-compose.yml` files that are not `services/*/docker-compose.yml` should go. I suspect docker-compose.yml in a service directory will be in support of automated testing -- we still need to decide what container orchestration will be used on Avena (Docker-compose, k8s, microk8s, k3s, kubeedge). However, I suspect testing will be done with compose for ease.

aaron97neu:
Agreed. Will split up the `docker-compose.yml` into each folder and push it tonight or tommorow after some testing

Looks like the command to run it will be `docker-compose -f folder1/docker-compose.yml -f folder2/docker-compose.yml ... up` which is pretty clunky but okay if it's just for testing I guess

abalmos:
Why not just `docker-compose up` ? I would expect each service to come with
a docker-compose that brings up whatever is needed to run automated tests
against it? These docker-compose files won't be useful in the real deploy
to Avena, at least that what the guy thinks.

For example: gps2tsdb should start a tsdb instance, gps2tsdb service, a
gpsd service. The test uses gpsd's simulator to send simulated GPS and then
checks the database.  Or at least something like that.

aaron97neu:
If you are in the `services` folder, `docker-compose up` will not work (it will fail to find a `docker-compose.yml`). If you cd in a certain service's directory, `docker-compose up` will work fine. This is okay if you are only running one service at a time but if you are doing multiple services at once (ex: `gps2tsdb` and `postgres` and `oada_upload` for a full GPS -> oada sync test), I believe you have to use `-f` to specify each `docker-compose.yml` file

abalmos:
I guess I'm expecting the docker compose file in a specific service to be
complete. It should bring up whatever container that it needs to have a
fully working flow.

aaron97neu:
Split up `docker-compose.yml` into each folder.

@abalmos I didn't have one for `gps2tsdb`. Was that perhaps lost somewhere? Will continue to look but since you developed that maybe you can find or recreate it faster?

PR #45, opened 2020-07-27
GPS to TimescaleDB in Python

aaron97neu:
Rewrote Andrew's GPS->TSDB in python using DBUS under Andrew's guidance

TODO:
- [x] Cleanup requirement and imports in `requirements.txt` and `Dockerfile`
- [x] Add code comments in `gps.py`
- [x] Multistage build


services/gps2tsdb/gps.py
```diff
@@ -0,0 +1,55 @@
...
+
+# Process GPS data, called every time a fix is acquired
+def fix(*args):
+    # Data comes in as an array, extract the relevant data. See the DBUS section of `man 8 gpsd` for more information
+    time = float(args[0])
+    lat = float(args[3])
+    lng = float(args[4])
+
+    # Print it all out for debugging
+    print("Time: ", time, "\tLat: ", lat, "\tLng: ", lng)
```
abalmos, line 23: We need to use some sort of ENV based logger (like `debug` for node.js). Everything that makes it to stdout is logged to the disk by docker.

services/gps2tsdb/gps.py
```diff
@@ -0,0 +1,55 @@
...
+    print("Inserting lat and lng for timestamp", time)
+    # Insert into database using global db var
+    db.run("INSERT INTO gps (time, lat, lng) VALUES (to_timestamp(%s), %s, %s)", (time, lat, lng) )
```
abalmos, line 26: When we merge this, can we also open a issue noting that we aren't logging all of the GPS data yet.

services/gps2tsdb/requirements.txt
```diff
@@ -0,0 +1,4 @@
+postgres
+pycairo
```
abalmos, line 2: isn't pycairo a graphical environment library?

services/oada_upload/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3.1'
 services:
   oada-upload:
     build:
-      context: ./
+      context: ../oada_upload
```
abalmos, line 5: I guess I see what your doing. I assume your trying to tie all these together so you `-f` them all together? 

Change my mind: docker-compose.yml files in a service should be a standard alone file that will bring up whatever is needed to test the service. This may be include things like `gpsfake` so `gps2tsdb.py` has something to work with. Without that, me thinks automated testing will be very hard.

services/gps2tsdb/requirements.txt
aaron97neu: Not sure but it does not seem vital. Will be removed with multistage build

services/gps2tsdb/gps.py
aaron97neu: Will do. Also should decrease the verbosity of the logging as well. 3+ lines per point is a bit much

services/gps2tsdb/docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+version: '3.1'
+services:
+  # docker run -it -v /var/run/dbus:/var/run/dbus --rm --name my-running-app my-python-app
```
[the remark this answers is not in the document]
aaron97neu: This is done in the volume section of the file

services/gps2tsdb/requirements.txt
aaron97neu: Not sure, but I believe it is not needed. Should be removed as part of the transition to multistage

aaron97neu: Not sure, but I don't think it is needed. It should be removed as part of the transition to multistage that I am working on

abalmos: I'm not sure that it will be easy to remove? I assume you will just copy the entire virtualenv directory into the final stage.

aaron97neu: You are correct... I removed `pycario` from the requirements list but is still seems gets built and copied over... Could be a dependency of `PyGObject`?

services/oada_upload/docker-compose.yml
aaron97neu: Agreed, however this is easy to change once we write an automated test. As of right now my testing procedure needs this

services/gps2tsdb/docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+  gps2tsdb:
+    build:
+      # Cannot use `./` due to this issue https://github.com/docker/compose/issues/3530#issuecomment-222490501
```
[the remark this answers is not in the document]
aaron97neu: Resolved to keep convo here: https://github.com/OATS-Group/isoblue-avena/pull/45#discussion_r461106242

services/gps2tsdb/docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+      context: ../gps2tsdb
+    restart: always
+    environment:
+      db_user: "avena"
+      db_host: "postgres"
+      db_database: "avena"
+      db_password: "password"
+      db_port: "5432"
+    depends_on:
+      - postgres
```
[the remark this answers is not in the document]
aaron97neu: The postgres/timescaledb docker container. I suppose this container could be run independently for testing purposes, however I am unsure how useful it will be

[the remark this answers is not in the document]
abalmos: I suppose my concern is that there is no way to know where to get that `postgres` service (and in this case it really is tsdb) because it is defined in a different docker-compose file that I have to remember to include. 

I am a bit confused on the point of these docker-compose files? A compose file is all about deploying a system somewhere. It is very unlikely _this_ compose file will end up on an Avena device.  An Avena device will need one customized to its needs. That why I assumed this compose file is about testing. Maybe I have misunderstood the goal?

Note -- I'm fine with bringing all this in for now as we work through what we want. This just seems like a nice moment to discuss what is it that we want.

services/gps2tsdb/requirements.txt
abalmos: Likely. DBUS started as a communication tool for use in graphical environments (which is why gdlib wrote a lib for dbus). SystemD made it popular outside of that with sd-bus. Maybe for the gdlib stuff to work it just has to bring in a bunch of the gdlib's other dependencies. In the long run run, we might want to look at Jeepney, a pure Python DBUS client (https://gitlab.com/takluyver/jeepney).  A bit more verbose API (but also less surprising).

services/gps2tsdb/Dockerfile
```diff
@@ -1,18 +1,34 @@
...
+# Needed for building PyGObject
+RUN apt-get install -y --no-install-recommends libgirepository1.0-dev gcc pkg-config python3-dev gir1.2-gtk-3.0
 
 COPY requirements.txt ./
-RUN pip install --no-cache-dir -r requirements.txt
+RUN pip install --user --no-cache-dir -r requirements.txt
+
+
+# RUNTIME
+FROM python:3 as runtime
```
abalmos, line 22: I think we settled on debian-slim. So maybe `python:3.8-buster-slim` or just `python:3-slim` depending on how locked down we want it to be? Or maybe I am just recalling my preference :rofl: With deploys over cell, I do think we want to be careful (but not absurd) on image size.

Interestingly it looks like `python` on dockerhub builds python from source rather than using Debian's python 3.8 https://github.com/docker-library/python/blob/44ded7c92b0c20abbeaa88785751eb0822d05ced/3.8/buster/slim/Dockerfile#L30

maybe that is to just move faster than Debian?

services/gps2tsdb/Dockerfile
```diff
@@ -1,18 +1,34 @@
...
+
+# Install runtime deps
+RUN apt-get -y update && apt-get install -y --no-install-recommends python3-dbus
```
abalmos, line 25: Should we `pip install` this since the python image does not use the Debian python package? ... I have no idea.

services/gps2tsdb/Dockerfile
abalmos, line 22: I think we settled on `debian-slim`. So maybe `python:3.8-buster-slim` or just `python:3-slim` depending on how locked down we want it to be? Or maybe I am just recalling my preference rofl With deploys over cell, I do think we want to be careful (but not absurd) on image size.

Interestingly it looks like python on dockerhub builds python from source rather than using Debian's python 3.8 https://github.com/docker-library/python/blob/44ded7c92b0c20abbeaa88785751eb0822d05ced/3.8/buster/slim/Dockerfile#L30

maybe that is to just move faster than Debian?

services/gps2tsdb/docker-compose.yml
abalmos, line 16: Maybe I'm not explaining what I assumed the docker-compose would be like. Not saying it should be this way, but maybe it will clarify earlier comments. That said, I can sort of see automated tests around something like this.

```suggestion
version: '3.1'
services:
  # docker-compose run --rm gps2tsdb bash run-tests.sh
  gps2tsdb:
    build:
      context: ./
    restart: never
    environment:
      db_user: "avena"
      db_host: "postgres"
      db_database: "avena"
      db_password: "password"
      db_port: "5432"
    depends_on:
      - tsdb
      - gpsd
    volumes:
      - dbus:/var/run/dbus
  tsdb:
    image: timescale/timescaledb:latest-pg12
    restart: never
    environment:
      POSTGRES_PASSWORD: "password"
      POSTGRES_USER: "avena"
      POSTGRES_DB: "avena"
      TIMESCALEDB_TELEMETRY: "off"
    ports:
      - 5432:5432
    volumes:
      - postgres:/var/lib/postgresql/data
      - ./postgres/schema:/docker-entrypoint-initdb.d
      - dbus:/var/run/dbus
  gpsd:
    image: <...some gpsd image ...>
    restart: never
    volumes:
      - dbus:/var/run/dbus
  gpsfake:
    image: <...some gpsfake image ...>
    restart: never
    depends_on:
      - gpsd
  sd-bus:
    image: <... some dbus-deamon image ...>
    restart: never
    volumes:
      - dbus:/var/run/dbus
volumes:
  postgres:
  dbus:
```

services/gps2tsdb/Dockerfile
aaron97neu: We never settled on something but it looks like we were actually leaning towards not slim https://groups.google.com/forum/#!msg/isoblue/ICwMrcDmk_w/IoZYFnwOBQAJ

Does not matter to me as long as we pick one.

services/gps2tsdb/docker-compose.yml
aaron97neu: Lets discuss this in the 7/29 meeting... I agree with you on what the `docker-compose.yml` file should be used for in the future, however I think we should not change it until we write the automated test. The way it is right now allows me to manually test everything and until we have an automated test ready to replace that lets not take that ability away

services/oada_upload/docker-compose.yml
aaron97neu: Moving discussion here https://github.com/OATS-Group/isoblue-avena/pull/45#discussion_r461692316

services/gps2tsdb/requirements.txt
aaron97neu: Agreed. I think we stick with it for now since it works and make an issue to transition to jeepney

services/gps2tsdb/docker-compose.yml
aaron97neu: A bit more clarification based on your comment here: https://github.com/OATS-Group/isoblue-avena/pull/45#discussion_r461653429

Right now the `docker-compose.yml` files are setup so I can do a `docker-compose -f xxx/docker-compose.yml -f yyy/docker-compose.yml ... up` to bring up the services I want for manual testing or deployment. In the (very very near) future we should transition to ansible/k3s/k8s for deployment and `docker-compose.yml` for automated testing

That is how I see it at least. Lets discuss in the 7/29 meeting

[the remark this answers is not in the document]
aaron97neu: Resolving to keep discussion here: https://github.com/OATS-Group/isoblue-avena/pull/45#discussion_r461692316

aaron97neu:
In an effort to get the ISOBlue location app working sooner rather than later, I am planning on merging and making the following issues for further enhancement unless any major problems are found during 7/29 meeting:
- Transition to a DBUS library that does not require 100s of MB of dependencies, (Consider https://gitlab.com/takluyver/jeepney, however it is still unstable)
- Further optimization of multistage build (transitioning to a better DBUS lib will help)
- Note we are not logging all GPS data
- Find and implement an ENV based logger like `debug` for node.js
- Automated tests need to be written and `docker-compose.yml` needs to be overhauled to support automated testing

I will keep tinkering with things and see if I can partially or fully address some of these before the meeting

services/gps2tsdb/docker-compose.yml
abalmos: If you remove the parts where we don't yet have an image made for it, it still works just like your current implementation except the `gps2tsdb` service is in charge of the postgres/timescale version, the schema, etc. and a developer just has to run `docker-compose --rm star gps2tsdb` to get the test images spun up. 

I don't think that's far from what we talked about how Avena may work. `gps2tsdb` would call on the timescale controller/manger to create it a user, database and it would establish the schema within that namespace.

abalmos:
> * Transition to a DBUS library that does not require 100s of MB of dependencies, (Consider https://gitlab.com/takluyver/jeepney, however it is still unstable)

A wrapper to make Jeepney a bit easier to use: https://github.com/ocaballeror/jeepney-objects/issues/2 . I linked to that issue because there are some enlightening comments on the state of Jeepney.

We could also just make a base image that comes with the libdbus stuff already in it and then base all of our python dbus services off of it -- so we only have to ship those bits once. pydbus has a pretty nice API but it appears dead. That said, DBUS isn't evolving either.  DBus support in Python is :-1:

PR #61, opened 2020-08-09
Container maintainer

### LINKED ISSUE #53: Deploying and Managing Containers
Currently we have no way to manage and deploy docker containers to our Avena devices in the field. In the few instances where I have needed to test containers in the field, I have manually `git clone` the repo and used `docker-compose` to start it manually. This is not elegant or sustainable. 

Possible solutions:
- [Ansible based](https://docs.ansible.com/ansible/latest/modules/docker_compose_module.html): Copy a Docker-Compose file over, possibly using a jinja2 template to configure it. Not elegant, and we are leaning away from using ansible for anything other than the initial deploy
- [k8s](https://kubernetes.io/)/[k3s](https://k3s.io/)/other kubernetes based approach - The Kubernetes model does not map to our model perfectly and would require us to use Kubernetes in a way it was not originally build for
- [BalenaEngine](https://www.balena.io/engine/) - We don't know much about this or if it is applicable, appears to be a souped up `docker-compose` wrapper. Seems more aimed at a docker replacement than a container management tool
- Homebrew custom solution - needs development effort that could be better spent elsewhere

@abalmos Are there any other approaches we discussed that I am missing?

aaron97neu:
Resolves #53 

TODOs:
- [x] Implement systemd with timers for periodic execution
- [x] Add Ansible deploy stuff
- [ ] ~Check for updates to the docker containers as well as the docker-compose file~
- [x] Check if `docker compose up` command fails (duh?)
- [x] When checking for changes to `docker-compose.yml`, use `docker-compose config` commands to generate the 'real' config files and check for equivalency (`docker-compose config` removes comments, expands relative filepaths to absolute, etc)
- [x] Use the docker python api when we can
- [x] Testing testing testing

Thoughts:
- Where should the script live? In `/usr/local/bin` like `can-watchdog`? It's not really a binary but there are already python scripts there...
  - Currently living in `/usr/local/bin/container-maintainer/`
- Do we want to containerize this? Unsure what interacting with containers while being a container would look like
  - Currently not containerized but an issue to containerize this should be made upon merge


aaron97neu:
Per meeting with @facastiblancor 8/17, we decided that the automatic updating of containers was unnecessary. Containers should be tagged with the desired version in the docker-compose file and updates will be done by modifying the docker-compose file (triggering a docker pull during the restart)

ansible/avena/roles/core/files/container-maintainer.py
```diff
@@ -0,0 +1,84 @@
...
+# 2. Use `docker-compose config` to generate 'validated' version of local config
+# 3. Check if generated docker-compose.yml files differnt
+# 4. If docker-compose.yml files differ, shut down all containers and restart them
+
+# Boolean to track if a container restart is needed
+restartcontainers=False
+
+# Get current docker-compose from the server
+hostname = socket.gethostname()
+url = 'http://172.16.0.1:8006/' + hostname + '/docker-compose.yml'
```
abalmos, line 22: Seems like we may want multiple docker-compose files? At least our core services and then one per "experiment" running. Maybe this is all v2, but it seems like "experiment" compose files should not have any enhanced privileges.

ansible/avena/roles/core/files/container-maintainer.py
```diff
@@ -0,0 +1,84 @@
...
+print('Querying server for docker-compose file')
+response = requests.get(url)
+
+# Ensure we got a valid response
+print('Ensuring we got a valid response')
+if response.status_code != 200:
+    print('Got a HTTP response code ', response, '. Is there a docker-compose.yml file in the right location?')
+    sys.exit()
```
abalmos, line 30: Will this leave ISOBlue dead in the field?

ansible/avena/roles/core/files/systemd/container-maintainer.timer
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Run container-maintainer.service every day
+Requires=container-maintainer.service
+
+[Timer]
+Unit=container-maintainer.service
+OnCalendar=daily
```
abalmos, line 7: Let's set a time of day so we know when it should have been done.

aaron97neu: Understood. Currently it is at midnight meaning that will be done right when the ISOBlue boots in the morning

ansible/avena/roles/core/files/container-maintainer.py
aaron97neu: No, it will use the last known good `docker-compose.yml` file it got from the server. If this is the first time it is running, there will be no docker containers running but the ISOBlue will still be accessible through SSH

ansible/avena/roles/core/files/container-maintainer.py
aaron97neu: Agreed. I think we consider this in v2

ansible/avena/roles/core/files/container-maintainer.py
```diff
@@ -0,0 +1,84 @@
...
+# Extract the docker-compose
+serverdc = response.content.decode('ascii')
+
+# Validate docker-compose
+open('./docker-compose.yml.new', 'w').write(serverdc)
+print('Validating new docker-compose')
+serverdc_config = subprocess.run(['docker-compose', '-f', './docker-compose.yml.new', 'config'], capture_output=True)
+
+if serverdc_config.returncode != 0:
+    print('docker-compose validation return code is not valid! Not using new docker-compose')
```
facastiblancor, line 41: If this check fails, will the maintainer fall back to the last known good `docker-compose.yml`?

ansible/avena/roles/core/tasks/docker.yml
```diff
@@ -82,3 +82,55 @@
   become: yes
   command:
     cmd: python3 -m pip install -IU docker-compose
+
+- name: container-maintainer
```
aaron97neu, line 86: This fails during a deploy from scratch due to the fact that wireguard has not been setup yet so it cannot connect and download the images. Not merging until I get it figured out.

First solution I though of was to just enable the service but not start it and let it run the next day and assume everything has been setup by then, but doesn't feel the best

aaron97neu:
Currently chasing an issue where the python script gets confused upon initial deploy but works fine after deleting the `docker-compose.yml` file or deleting all container-maintainer related files and rerunning ansible. Quite strange honestly

@abalmos I just pushed a pretty significant ansible reorganization (pulling the docker stuff out of core to ensure it can be run after wireguard). Manual testing looks okay, but could you glance over it and make sure things look alright?

PR #62, opened 2020-08-26
Received cell signal strength logger

### LINKED ISSUE #33: Create an app to log received cell signal power to a CSV file.
An app that polls the RSSI power data from ModemManager.

facastiblancor:
Containerized version of cell-logger. Runs in a container and uses DBus to get data from the ModemManager and saves it to `/data/log/cell.csv.` Closes #33 

TO DO:

- [ ] Test app stability across sleep/reboot cycles
- [x] Add postgres database logging.


services/cell-logger/cell-logger
```diff
@@ -0,0 +1,72 @@
...
+    modem_signal_iface = 'org.freedesktop.ModemManager1.Modem.Signal'
+
+    modem_manager_object = bus.get_object('org.freedesktop.ModemManager1', 
+                                          '/org/freedesktop/ModemManager1')
+    modem_manager_iface = dbus.Interface(modem_manager_object, 
+                                         'org.freedesktop.DBus.ObjectManager')
+
+    modem_data = modem_manager_iface.GetManagedObjects()
+    
+    modem_data_str = str(modem_manager_iface.GetManagedObjects())
```
aaron97neu, line 21: Is this variable used anywhere? Remove it if not

facastiblancor: It was used to store and then print debug information. Now it went unused and will be removed.

.github/workflows/main.yml
```diff
@@ -24,6 +24,7 @@ jobs:
         service:
           - gps2tsdb
           - oada_upload
+          - cell-logger
```
abalmos, line 27: I think this has to be `cell_logger` ... I can't remember what something chokes on the `-`'s

services/cell-logger/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN apt-get -y update
+
+# Needed for building dbus-python
+RUN apt-get install -y --no-install-recommends python-pip pkg-config libdbus-1-dev libglib2.0-dev libdbus-glib-1-dev python3-dbus
+
+# Activate virtualenv
+RUN python -m venv /opt/venv
+
+# Make sure we use the virtualenv
+ENV PATH="/opt/venv/bin:$PATH"
```
abalmos, line 16: Seems like it might be time to factor this out into a base container. Maybe I can make that a todo for me after the PR is merged.

services/cell-logger/cell-logger
```diff
@@ -0,0 +1,85 @@
...
+
+while(True):
+
+    timestamp = int(time.time()) 
+
+    [signal, cell_tech] = get_modem_rssi()
+    
+    if (signal != ''):
+
+        with open('/data/log/cell.csv', mode = 'a') as log:
```
abalmos, line 77: Does this `fopen` and `fclose` for every loop?

services/cell-logger/cell-logger
```diff
@@ -0,0 +1,85 @@
...
+
+            log = csv.writer(log, delimiter = ',', quotechar = '"',
+                             quoting = csv.QUOTE_MINIMAL)
+
+            log.writerow([timestamp, signal, cell_tech])
+            print(timestamp, signal, cell_tech)
+
+    time.sleep(1)
```
abalmos, line 85: Okay, but does Python have a real timer that can be used to schedule events? Will think block a kill signal for one second?

aaron97neu: `sleep()` appears to respond to `Ctrl-C`

```
Python 3.8.5 (default, Aug  2 2020, 15:09:07) 
[GCC 10.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from datetime import datetime
>>> from time import sleep
>>> print(datetime.now())
2020-08-27 16:38:15.503645
>>> sleep(9999999)
^CTraceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyboardInterrupt
>>> print(datetime.now())
2020-08-27 16:38:25.319574
```

services/cell-logger/cell-logger
facastiblancor: Yes, every time the code inside the _with_ statement is ran, all the resources used will be closed. This is likely to change (and eventually disappear) after a transition to a local postgres database logging.

.github/workflows/main.yml
aaron97neu: @facastiblancor probably should change this.. will need to rename all instances of `cell-logger` to `cell_logger` like https://github.com/OATS-Group/isoblue-avena/pull/63/commits/3abf9024df7fc3a55229cfd492f04fe54c7480f2

PR #63, opened 2020-08-27
Containerization of can-watchdog

### LINKED ISSUE #51: canwatchdog to container
# Implementation idea
1. base debian docker image with watchdog script
2. use dbus to issue a `systemd sleep` call
3. use normal canutils to bring up can interfaces --- web service to configure that ... maybe this should be part of the "controller" service.

# Gotchas
1. Need use to host networking mode to access can interfaces

### LINKED ISSUE #59: CAN Watchdog "Stay Online" Flag
From meeting with @aultac on 8/7/2020

There should be some sort of flag or mechanism to ask the ISOBlue to stay online even if no CAN is detected. This would be very helpful to keep demos and deploys from being interrupted.

There are several ways we could do this, however containerizing `can-watchdog` may add some complications. Would using DBUS be feasible? 

This likely should be done in conjunction with #51

aaron97neu:
Containerizes the previously standalone sleep-without-can functionality of Avena. Closes #59 and Closes #51

TODOS:
- [x] Transition from `systemctl suspend` to dbus
- [x] Test test test
- [ ] Implement ability to disable suspend in certain situations (which? How?)


.github/workflows/main.yml
```diff
@@ -24,6 +24,7 @@ jobs:
         service:
           - gps2tsdb
           - oada_upload
+          - can-watchdog
```
abalmos, line 27: I don't recall what, but something chokes on `-` in the name

services/can-watchdog/can-watchdog.py
```diff
@@ -0,0 +1,26 @@
...
+print(len(rx_paths), ' found: ', rx_paths)
+
+prev_rx = 0
+curr_rx = 0
+
+sleep(5)
+
+while True:
+    for rx in rx_paths
+        with open(rx, 'r') as file:
```
abalmos, line 18: Does this `fopen` and `fclose` for each loop?

services/can-watchdog/can-watchdog.py
```diff
@@ -0,0 +1,26 @@
...
+print('Getting all can rx counters')
+rx_paths = glob.glob('sys/class/net/can*/statistics/rx_bytes')
+print(len(rx_paths), ' found: ', rx_paths)
+
+prev_rx = 0
+curr_rx = 0
+
+sleep(5)
+
+while True:
```
abalmos, line 16: Shall we delay between each loop?

aaron97neu: .... yes

services/can-watchdog/can-watchdog.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/python3
+import glob # For expanding wildcard filepaths
+import jeepney # For interactions with dbus
+from subprocess import call
+
+print('CAN Watchtog has started')
+print('Getting all can rx counters')
+rx_paths = glob.glob('sys/class/net/can*/statistics/rx_bytes')
```
abalmos, line 8: I'm not sure how to fix, but this will break if we were to name the can interfaces something other then can*.

Could maybe query netlink directly and find all can interfaces to check. libnl looks useful. Not sure though.

services/can-watchdog/can-watchdog.py
aaron97neu: Fixed

services/can_watchdog/can_watchdog.py
```diff
@@ -0,0 +1,94 @@
...
+curr_rx = 0
+
+rx_fds = []
+for rx in rx_paths:
+    rx_fds.append(open(rx, 'r'))
+
+print('Setup complete. Beginning rx_bytes watch')
+while True:
+    print('Sleeping 5')
+    sleep(5)
```
abalmos, line 75: Should we make this an ENV so we can adjust the inactivity threshold? I think it needs to be at least long enough that you can login with the shell and bring the container down. Seems like 5 seconds is too short to type username, password, and paste + enter the `cd <blah> && docker-compose stop` line. This is mostly a in-the-field-and-its-broke type situation.

abalmos:
Not for this time around but I wanted to write down that it may also be useful to allow the other tools a chance to do something before suspend occurs. For example: flush the disks, send a message to the server, etc. etc. Another potential role for the admin container.

I suppose this similar to having a systemd service depend on systemd-suspend.service, etc.

Issue #65, opened 2020-08-31
Bring Prometheus/Grafana back 


abalmos:
Restore the Prometheus/Grafana dashboard functionality. It was lost during move to oats2 and new docker-compose strategy.

aaron97neu:
@abalmos What still needs to be done to close this issue/

abalmos:
Nothing. This work requires no PRs to `isoblue-avena`. One day, we should have some example compose files for `container-maintainer` and then we could include `node-exporter`, `cadvisor`, etc.

Setting up prometheus is not dependent on Avena. We could show some example configs and a dashboards though?

PR #71, opened 2020-09-18
Add gps prometheus metrics

abalmos:
Adds Prometheus GPS fix metrics to gps2tsdb

### DIFF
--- a/services/gps2tsdb/gps.py
+++ b/services/gps2tsdb/gps.py
@@ -18,11 +18,11 @@ def fix(*args):
     print("Finsihed inserting lat and lng for timestamp ", args[0])
 
 global lat_gauge
-lat_gauge = Gauge('position_lat', 'Last know fix latitude')
+lat_gauge = Gauge('avena_position_lat', 'Last know fix latitude')
 global lng_gauge
-lng_gauge = Gauge('position_lng', 'Last know fix longitude')
+lng_gauge = Gauge('avena_position_lng', 'Last know fix longitude')
 global time_gauge
-time_gauge = Gauge('position_time', 'Last know fix time')
+time_gauge = Gauge('avena_position_time', 'Last know fix time')
 start_http_server(10001)
 
 global db
--- a/services/gps2tsdb/gps.py
+++ b/services/gps2tsdb/gps.py
@@ -4,14 +4,26 @@ from dbus.mainloop.glib import DBusGMainLoop
 import dbus
 import postgres
 import os
+from prometheus_client import start_http_server, Gauge
 
 def fix(*args):
     #print(args)
     print("Time: ", args[0], "\tLat: ", args[3], "\tLng: ", args[4])
+    lat_gauge.set(args[3])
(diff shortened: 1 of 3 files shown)


abalmos:
This is running on dev5, dev6, and 4. Appears to work. Will be on 2 on next upgrade.

Note: Need to fix dev5, dev,6 4, and 2 to `isoblue/gps2tsdb:latest` after merge.

aaron97neu (approved):
LGTM and we have seen the data appear on the grafana dashboard. Merge whenever ready

aaron97neu:
Will this close #65?

services/gps2tsdb/gps.py
```diff
@@ -4,14 +4,26 @@
...
     db.run("INSERT INTO gps (time, lat, lng) VALUES (to_timestamp(%s), %s, %s)", (float(args[0]), float(args[3]), float(args[4]) ) )
     print("Finsihed inserting lat and lng for timestamp ", args[0])
 
+global lat_gauge
+lat_gauge = Gauge('position_lat', 'Last know fix latitude')
+global lng_gauge
+lng_gauge = Gauge('position_lng', 'Last know fix longitude')
+global time_gauge
+time_gauge = Gauge('position_time', 'Last know fix time')
+start_http_server(10001)
```
abalmos, line 26: @aaron97neu Is this `global` business the Python way? I didn't see a clear why to do it otherwise.

abalmos:
> Will this close #65?

No, but #65 is essentially done. Committing the new container maintainer compose files to `oats-fleet` would be the final task for that.

services/gps2tsdb/gps.py
aaron97neu: I was actually thinking about this... [PEP8](https://www.python.org/dev/peps/pep-0008/) does not mention anything about using globals. I am far from an expert on things being "pythonic" but I think they should be avoided similar to other languages. However, I am not sure how to pass that information to `fix()` unless `bus.add_signal_receiver()` allows us to specify additional parameters...

One alternative way would be to specify a single 'settings' global dictionary (Or 1 for db stuff and 1 for prometheus). The information would be accessed via `prometheus_config['lat_gauge']`, `prometheus_config['lng_gauge']`, etc. This would avoid a bunch of globals floating around, make adding additional prometheus data points easier (altitude, accuracy, etc), and if we do find a way to pass additional parameters to `fix`, we would only need one additional parameter

At the end of the day the script is readable (but could use some comments) and works. In the future I expect we will rewrite this with the Jeepney lib and it may not be a problem then. I vote we leave it

abalmos: Fair points. Seems like a tracing lib like prometheus would have to rely on globals or just about everything would have to accept it as a parameter. 

Feels less weird to pass the `db` instance around because only certain functions need to use the DB -- but like you said, that does not appear to be possible with this DBUS setup.

In the re-write, reducing to a one or two global maps would at least be an improvement.

PR #72, opened 2020-09-18
Add oats2 ga runner

abalmos:


abalmos:
This works now. Some downsides. Lets evaluate on next call.

PR #75, opened 2020-09-21
Prometheus client metric export for cell logger.

### LINKED ISSUE #73: Add Prometheus metrics to cell_logger
**Use prometheus metrics in _`cell_logger`_ to visualize the RSSI in Grafana.** 

The plan is to add a prometheus client to the current `cell_logger` script, so the information gathered by cell_logger can be visualized in a Grafana dashboard.

facastiblancor:
This adds a received signal metric export using prometheus. This is to add the ability to view the received cell signal power in a Grafana dashboard. Closes #73 .

### DIFF
--- a/services/cell_logger/cell_logger.py
+++ b/services/cell_logger/cell_logger.py
@@ -5,6 +5,7 @@ import time
 import csv
 import postgres
 import os
+from prometheus_client import start_http_server, Gauge
 
 #Query modem data using the ModemManager DBus object
 
@@ -81,6 +82,11 @@ def write_to_db(timestamp, signal, cell_tech):
 
     print("Finished inserting cell signal power data for timestamp ", timestamp)
 
+
+#Create a metric to track cell signal
+
+signal_gauge = Gauge('received_cell_signal_power', 'Received cell signal power')
+
 #Initialize postgres database
 
 global db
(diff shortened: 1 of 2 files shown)


abalmos (approved):
LGTM.

I say merge when once Grafana can plot the data.

I would like to run this on the WiFi boxes as well. I'm not sure if the self-runner stuff I setup will break a DockerHub deploy. Let me know if it does.

services/cell_logger/cell_logger.py
```diff
@@ -81,6 +82,11 @@ def write_to_db(timestamp, signal, cell_tech):
 
     print("Finished inserting cell signal power data for timestamp ", timestamp)
 
+
+#Create a metric to track cell signal
+
+signal_gauge = Gauge('received_cell_signal_power', 'Received cell signal power')
```
abalmos, line 88: I would change the name of this metric. Things like to break metrics names by `_` so the first word because a "group name" for a set of metrics.

Maybe `cell_receive_power` ? Then we can have things like `cell_access_type` , `cell_sms_rx`, etc. in the future.

abalmos: Maybe even, `mm_` rather then `cell_`?

abalmos (changes requested):

services/cell_logger/cell_logger.py
facastiblancor: I left `avena_cell` to indicate that this cell signal comes from a component of the Avena stack as discussed.

abalmos (approved):

PR #77, opened 2020-09-22
CSV/DB selection for cell logger.

facastiblancor:
This uses environment variables to select between CSV and DB logging in cell-logger. The environment variables should be specified in docker-compose.

### DIFF
--- a/services/cell_logger/cell_logger.py
+++ b/services/cell_logger/cell_logger.py
@@ -95,6 +95,8 @@ signal_gauge = Gauge('avena_cell_signal_power', 'Received cell signal power')
 
 if((log_env == 'DB') or (log_env == 'DB,CSV') or (log_env == 'CSV,DB')):
 
+    import postgres
+
     global db
 
     connection_url = ('postgresql://' + os.environ['db_user'] + ':' + 
@@ -119,7 +121,7 @@ if((log_env == 'DB') or (log_env == 'DB,CSV') or (log_env == 'CSV,DB')):
             migrate_data => TRUE);""")
 
     print("Finished setting up tables")
- 
+
 #Initialize DBus system bus
 
 bus = dbus.SystemBus()
--- a/services/cell_logger/cell_logger.py
+++ b/services/cell_logger/cell_logger.py
@@ -3,7 +3,6 @@
 import dbus
 import time
 import csv
-import postgres
 import os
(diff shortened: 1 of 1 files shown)


facastiblancor:
Testing has shown good results. I will merge this evening if no issues arise.

services/cell_logger/cell_logger.py
```diff
@@ -82,38 +81,47 @@ def write_to_db(timestamp, signal, cell_tech):
...
 
 #Create a metric to track cell signal
 
 signal_gauge = Gauge('avena_cell_signal_power', 'Received cell signal power')
 
-#Initialize postgres database
+#Initialize postgres database if "CELL_LOG" environment variable is set to log
+#to database.
+
+if((log_env == 'DB') or (log_env == 'DB,CSV') or (log_env == 'CSV,DB')):
```
abalmos, line 96: This seems fragile, but working. I say if we extend this anymore, we should look for a different solution.

services/cell_logger/cell_logger.py
```diff
@@ -130,8 +138,14 @@ def write_to_db(timestamp, signal, cell_tech):
     
     if (signal != ''):
         
-        write_to_csv(timestamp, signal, cell_tech)
-        write_to_db(timestamp, signal, cell_tech)
+        if((log_env == 'CSV') or (log_env == 'CSV,DB') or (log_env  == 'DB,CSV')):
```
abalmos, line 141: And too repetitive. 

This could help:  https://pypi.org/project/environs/

Something like this might be best in the future: https://click.palletsprojects.com/en/7.x/options/#values-from-environment-variables

That allows config through both CLI and ENV. Could something like:

./cell_logger --csv <path> --postgres <connection_url>

AND

CSV=<path> POSTGRES=<connection_url> ./cell_logger

with the same code

abalmos (approved):
Using on WiFi boxes. Seems to work. Grafana has troubles with the data sometimes, but I think that is Grafana.

Left some notes fohttps://click.palletsprojects.com/en/7.x/options/#values-from-environment-variablesr next time.

:shipit:

services/cell_logger/cell_logger.py
facastiblancor: I agree. Something better could be done. Better to fix now than later in a rush.

PR #79, opened 2020-09-28
Removal of old kafka related containers

aaron97neu:
As discussed last week, there are a few containers in our services directory that are no longer in use. They have been removed.

Not sure if a bug or feature but it does not add any new functionality for went with bugfix.


abalmos (approved):
LGTM :shipit:

Issue #70, opened 2020-09-18
postgres triggers should start with service name


abalmos:
I suggest we name postgres triggers something like:

`avena_[service]_[function_name]`

For example: `create_sent_row_trig` -> `avena_oada_upload_create_sent_row`

aaron97neu:
Should have been closed with #80

PR #80, opened 2020-09-28
oada_upload index and renaming sent triggers

### LINKED ISSUE #69: oada_upload postgres select query slow
Might need to make an index on (sent, timestamp) or something. Or maybe use `sent` as the space index in TSDB.

aaron97neu:
Adding a trigger to the `gps_sent` table. When testing on a `gps_sent` table of 3.5 million rows, this takes the load from postgres/tsdb from 0.6 to less than 0.03. Also renamed the triggers and functions while I was in there

Closes #69 and #70

Tested briefly on dev05. Requires a bit more testing, opening PR to being docker build process.

Plan is to test on dev06, then one prod machine, and then merge if no problems arise.

### DIFF
--- a/services/oada_upload/src/oada_upload.ts
+++ b/services/oada_upload/src/oada_upload.ts
@@ -188,6 +188,15 @@ async function main(): Promise<void> {
     $$;
   `);
 
+  // Remove old triggers and procedures if they exist
+  console.log(`Removing old triggers and functions`);
+  await db.query(`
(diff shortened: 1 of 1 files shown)


abalmos (approved):
LGTM :shipit:

aaron97neu:
Testing on dev06 looks good, awaiting prod machine to come online tomorrow morning

PR #83, opened 2020-09-30
CAN mode and bitrate configuration in Ansible

facastiblancor:
This PR adds bitrate configuration for each CAN interface. This helps to properly configure the bitrate, specially in John Deere machines, where implement and tractor bus run at different speeds. Also adds the option to enable _listen-only_ mode to avoid locking the CAN bus when the bitrate is set to a different value.


abalmos:
Are the changes to cell logger intended?

PR #84, opened 2020-09-30
Socketcand based CAN logger.

facastiblancor:
This PR adds a socketcand container to expose CAN bus data to a socket in the Docker network. Also includes _can_logger_, logging data obtained from socketcand and logging to a CSV file or a Database.


abalmos:
@facastiblancor I  have not had time to really look at this PR, but from the file listing I am concerned that we duplicating socketcand's code here, making getting upgrades / bug fixes difficult.

Should the dockerfile just pull a certain commit of socketcand in the build phase?

Did you mean to include changes to `cell_logger`?

aaron97neu:
@facastiblancor Looks like you branched off of `bugfix/can-configuration` not master. NBD if we merge that branch in first, @abalmos might have a more elegant idea on how to fix that. 

Agree about duplicating socketcand's code. Makes it hard to maintain and control. Would like to see pulling the commit via dockerfile or a git submodule

services/can_logger/can_logger.py
```diff
@@ -0,0 +1,95 @@
...
+
+import socket
+import postgres
+import csv
+import os
+
+def write_to_csv(timestamp, can_id, can_data, host_interface):
+
+
+    with open('/data/log/' + host_interface + '.csv', mode = 'a') as log:
```
aaron97neu, line 11: Curious about the performance implications of opening and closing the same file for each can message (many per second) received. While less elegant, a `csv_init()` method that returns a file descriptor that gets passed to `write_to_csv()` might be more performant. Might have to flush the fd here and there?

services/can_logger/can_logger.py
```diff
@@ -0,0 +1,95 @@
...
+
+        log = csv.writer(log, delimiter = ',', quotechar = '"',
```
aaron97neu, line 13: `csv_init()` might also want to make the `csv.writer()` object and pass that here instead. I am worried about calling `write_to_csv()` as is many times a second but I could be wrong

services/can_logger/can_logger.py
```diff
@@ -0,0 +1,95 @@
...
+while(True):
+    
+    frame = s.recv(128).decode("utf-8").strip("<>''").split(' ')
+    [timestamp, can_id, can_data] = (frame[3], frame[2], frame[4])
+    
+    if(logging.find('db') != -1):
+        
+        write_to_db(timestamp, can_id, can_data)
+    
+    else:
```
aaron97neu, line 93: What if we don't want to write to a csv? IE `log: "db"` in the docker-compose

services/can_logger/can_logger.py
```diff
@@ -0,0 +1,95 @@
...
+                         quoting = csv.QUOTE_MINIMAL)
+
+        log.writerow([timestamp, can_id, can_data])
+
+def write_to_db(timestamp, can_id, can_data):
+
+    db.run("INSERT INTO cell (time, can_id, can_data) VALUES (to_timestamp(%s), \
+            %s, %s)", (timestamp, can_id, can_data))
+
+    print("Finished inserting CAN bus data for timestamp ", timestamp)
```
aaron97neu, line 23: Should either remove this or also put a print statement in `write_to_csv()`. Prob want to remove, a print per CAN msg will clog up stdout very quickly

services/can_logger/can_logger.py
```diff
@@ -0,0 +1,95 @@
...
+
+def db_init():
+    
+    global db
```
aaron97neu, line 27: Returning a db var that is then passed to `write_to_db()` will likely be better practice. Should avoid globals if we can

services/can_logger/can_logger.py
```diff
@@ -0,0 +1,95 @@
...
+    
+    db_init()
+
+#Initialize host socket
+
+s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+
+s.connect((host_ip, host_port))
+
+print (s.recv(32))
```
aaron97neu, line 72: What do these print statements look like? Might want to add some sort of context if it isn't in whatever `s.recv(32)` gives us.  IE `print("init msg from socketcan:", s.recv(32))`

Some comments explaining why you are printing the first 96 bytes instead of logging them would be helpful

services/can_logger/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+    build:
+      context: ../can_logger
+    restart: always
+    environment:
+      db_user: "avena"
+      db_host: "postgres"
+      db_database: "avena"
+      db_password: "password"
+      db_port: "5432"
+      host_ip: "172.17.0.1"
```
aaron97neu, line 13: From `can_logger.py`:
```
host_ip = os.environ['socketcand_ip']
host_port = os.environ['socketcand_port']
host_interface = os.environ['can_interface']
logging = os.environ['log']
```

I think you mixed up the names somewhere? IE It should be `socketcand_ip: "xx.yy.zz.aa"` here or `host_ip = os.environ['host_ip']` there

services/can_logger/can_logger.py
```diff
@@ -0,0 +1,95 @@
...
+
+s.sendall(b'< rawmode >')
+
+print(s.recv(32))
+
+while(True):
+    
+    frame = s.recv(128).decode("utf-8").strip("<>''").split(' ')
+    [timestamp, can_id, can_data] = (frame[3], frame[2], frame[4])
```
aaron97neu, line 87: What is in `frame[0]`, `frame[1]`, etc? Are they useful or something we want to log?

facastiblancor:
> @facastiblancor I have not had time to really look at this PR, but from the file listing I am concerned that we are duplicating socketcand's code here, making upgrades / bug fixes difficult.
> 
> Should the dockerfile just pull a certain commit of socketcand in the build phase?
> 
> Did you mean to include changes to `cell_logger`?

Yeah, I'm not entirely confident of copying socketcand's code. If I pull the code using git, then the container would be bloating with lots of dependencies, increasing image size. This was an effort to keep the image size small. Other option could be bundling a binary instead of the source, but that would be way worse.

@aaron97neu  Didn't notice that. That branch is almost ready to merge, however this still needs testing with a full deploy.

services/can_logger/can_logger.py
```diff
@@ -0,0 +1,95 @@
...
+
+    print("Finished setting up tables")
+
+
+#Get host info using environment variables
+
+host_ip = os.environ['socketcand_ip']
+host_port = os.environ['socketcand_port']
+host_interface = os.environ['can_interface']
+logging = os.environ['log']
```
aaron97neu, line 58: See comment in `docker-compose.yml`. I think you mixed up the enviromental var names

aaron97neu (changes requested):
Also add these containers in `.github/workflows/main.yml`

aaron97neu:
> Yeah, I'm not entirely confident of copying socketcand's code. If I pull the code using git, then the container would be bloating with lots of dependencies, increasing image size. This was an effort to keep the image size small. Other option could be bundling a binary instead of the source, but that would be way worse.

If files you included are only the ones necessary for what we need, not the whole repo, you may have a point. Curious how crazy these dependencies are. Where did you get those files?

abalmos:
> Yeah, I'm not entirely confident of copying socketcand's code. If I pull the code using git, then the container would be bloating with lots of dependencies, increasing image size. This was an effort to keep the image size small. Other option could be bundling a binary instead of the source, but that would be way worse.

Bundling the binary isn't possible because we need to support multiple architectures, but it is _sort_ of the answer.

You should use a multistage Dockerfile, like we do with the python setup. In if the first stage you would `git clone` socketcand code, install compilers, libraries, etc. Run configure and make.
Then in the second stage you copy the resulting binary and install any libraries that socketcand is dynamic linked to (if any). The first stage could be just normal `debian` and the second stage `debian-slim`. I want to say that debian even has a build that comes with `gcc`, `make`, etc. for the first stage.

Doing it like this, socketcand code comes from the socketcand repo each time (could be a build arg to determine commit/tag), the binary is built for the current architecture (github actions ensures we build for both arm and amd64), and the containers are minimal (the first stage is thrown away after the build)

abalmos:
One other piece of feedback that could cause major structural changes: I don't think we want to hit postgres with 100s of INSERTs per second. I think can_logger will need to buffer up and do larger inserts.

Something like `every N messages or S seconds`?

A few inserts a second seems reasonable? @aaron97neu ?

aaron97neu:
> A few inserts a second seems reasonable? @aaron97neu ?

100s inserts/sec maybe, prob not. Would need to see testing. 

1s inserts/sec probably, testing still a good idea

I think inserts at a constant rate is a good idea. Implementation would be interesting. Maybe a queue that the main function writes arrays containing can frames to and a write_to_csv/db reads it regularly. Might be a good place to use multithreading or multiprocessing, the later being truly concurrent in cpython

services/can_logger/can_logger.py
facastiblancor: Not sure about implications with numbers, but with something as busy as CAN, it can heavily stress the CPU. My only concern is the moment to flush the file, as the logger is meant to always run, only to stop if an exception arises or the process is terminated.

services/can_logger/can_logger.py
facastiblancor: The first two items of `frame` are blank and space characters coming from socketcand's stdout strings.

services/can_logger/can_logger.py
aaron97neu: Not sure I follow. Could you explain a little more? Do you mean like the isoblue being unplugged in the middle of a flush?

services/can_logger/can_logger.py
facastiblancor: Indeed. This could be written more efficiently.

services/can_logger/can_logger.py
facastiblancor: It only prints socketcand's reply messages. These statements were left on debug purposes and are meant to be deleted as soon as code stabilizes.

services/can_logger/can_logger.py
facastiblancor: No, I mean, I am not sure when the fd can be flushed as the program is meant to always run and write to file. I understand that the fd should be flushed as soon as the program is done accessing the file, but in this situation, the program would be done only if it is stopped by the user. Other events might not be predictable, as unplugging or the program crashing for unknown reasons. I could be wrong though.

abalmos:
Is this ready for review?

facastiblancor:
> Is this ready for review?

Yes, most of it. However I _think_ it would be better to implement a separate thread to write to the DB. That's what is missing in the last commit.

services/can_logger/can_logger.py
```diff
@@ -0,0 +1,123 @@
...
+    log = open('/data/log/' + host_interface + '.csv', mode = 'a') 
+
+    return log
+
+def write_to_csv(timestamp, can_id, can_data, log):
+
+        log = csv.writer(log, delimiter = ',', quotechar = '"',
+                         quoting = csv.QUOTE_MINIMAL)
+
+        log.writerow([timestamp, can_id, can_data])
```
aaron97neu, line 19: Still not a huge fan of how it's being done. I _think_ this will work without modification of the main function. This will keep us from creating a new `csv.writer` object every time `write_to_csv()` is called


```suggestion
def csv_init(host_interface):
    # Open the log file
    logfd = open('/data/log/' + host_interface + '.csv', mode = 'a') 
    # Make csv writer object using log file
    logcsv = csv.writer(log, delimiter = ',', quotechar = '"',
                         quoting = csv.QUOTE_MINIMAL)
    return logcsv

def write_to_csv(timestamp, can_id, can_data, log):
        log.writerow([timestamp, can_id, can_data])
```

Also for some reason I feel like the `log` object should be the first argument. Not sure why, maybe a old C convention that the pass by reference arguments are usually first? @abalmos might have an opinion

services/can_logger/can_logger.py
```diff
@@ -0,0 +1,123 @@
...
+
+if(logging.find('db') != -1):
+    
+    db = db_init()
+
+# Initialize CSV file if CSV logging is enabled
+
+if(logging.find('csv') != -1):
+
+    csv_log = csv_init()
```
aaron97neu, line 73: Another nitpick, but it might be better to do something like this:

```suggestion
logtodb = False
logtocsv = False
if logging.find('db') != -1:
    logtodb = True
if logging.find('csv') != -1:
    logtocsv = True

# Initialize postgres database if database logging is enabled
if logtodb:
    db = db_init()

# Initialize CSV file if CSV logging is enabled
if logtocsv:
    csv_log = csv_init()
```

This way if we ever choose to change how to determine if we are doing csv or db logging, it's only one part of the code to change, and we can just check a boolean value instead of calling `.find()`, which might make a difference in the time sensitive `while(true)` loop below.

services/can_logger/can_logger.py
```diff
@@ -0,0 +1,123 @@
...
+
+        if(logging.find('db') != -1):
+        
+            # TODO: Write to database in a separate thread/process.
+            wr_buff = rx_buff.copy()
+            write_to_db(wr_buff)
+    
+    if(logging.find('csv') != -1):
+        
+        write_to_csv(timestamp, can_id, can_data, csv_log)
```
aaron97neu, line 123: In the case that there is only csv streaming, you are setting up and filling `rx_buff` and then never using it . Either rewrite `write_to_csv()` to use `rx_buff` (ideal since you will call `write_to_csv()` less) or put everything from `rx_buff.append...` until `write_to_db....` into a `if logtodb: ` block

.github/workflows/main.yml
```diff
@@ -26,6 +26,7 @@ jobs:
           - oada_upload
           - cell_logger
           - can_watchdog
+          - can_logger
```
aaron97neu, line 29:
```suggestion
          - can_logger  
          - socketcand
```

right?

services/can_logger/can_logger.py
facastiblancor: You mean, writing to CSV as a funcion instead of using the code in the main function? I thought it would be easier to understand if the code inside the loop has less amount of lines.

.github/workflows/main.yml
facastiblancor: Agree.

aaron97neu (approved):
Make an issue to to move `can_logger` to a process pool instead of creating a process for every 100 can messages. Other than that :shipit:

PR #85, opened 2020-09-30
Container-maintainer ping and pull hardening

aaron97neu:
I've been noticing that container-maintainer fails sometimes because it can't reach 172.16.0.1. I believe this is because the script is being run at startup and wireguard has not finished starting up yet. Made a few changes to solve this and overall harden container-maintainer:

Ping the server several times before failing to get docker-compose.yml.
Even if server with docker-compose.yml server is not availble, try a docker-compose pull

### DIFF
--- a/ansible/avena/roles/docker/files/container-maintainer.py
+++ b/ansible/avena/roles/docker/files/container-maintainer.py
@@ -1,10 +1,11 @@
 #!/usr/bin/env python3
 
-import requests   # Used to get docker-compose from server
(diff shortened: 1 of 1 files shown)


aaron97neu:
Diff is real strange in this one. Changes really only in the first 30 lines, plus some print statements and putting the `docker-compose.yml` get in an `if` block which might be what is confusing diff

abalmos (approved):
LGTM :shipit:

PR #86, opened 2020-10-01
Updates to Github Actions for release

abalmos:
This PR contains

1. major version upgrades of the underlying docker github actions. 

It is all much nicer now :tada: 

2. updates to the tagging scheme.

I *think* this is right:

- Things pushed to a branch like `release/<version>` are tagged `:<version>`, `:<major>`, `:<minor>`, and `:latest`. Where `<major>` is 1 and `<minor>` is 1.3 for a `<version>` of `1.3.24`.

- Things pushed to `master` branch (or really the GitHub "default branch" setting) are tagged with `:next`.

- Things pushed to any other branch are tagged with `:<branch-name>` where `<branch-name>` is the git branch name with `/` replaced by `-`. 

- Pull requests are tagged with `:pr-<pr-num>` where `<pr-num>` is the GitHub pull request number

- All pushes are ALSO tagged with `sha-<short-sha>` where `<short-sha>` is the first 8 characters of the git commit SHA.

With this setup, you can track a version (full, major, or minor), a branch, the current next-to-be-released, or a specific commit/push. 

These rules are simple to change (just a bash script). So we can discuss and modify as needed.

3. Added dependabot to notify us of action updates

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,8 +54,8 @@ jobs:
 
           # Release version is just the release number
           if [[ $GITHUB_REF == refs/heads/release/* ]]; then
-            VERSION=${GITHUB_REF#refs/heads/release/}
-            RELEASE=true
+              VERSION=${GITHUB_REF#refs/heads/release/}
+              RELEASE=true
 
           # Branch version is branch name (with '/' -> '-')
           elif [[ $GITHUB_REF == refs/heads/* ]]; then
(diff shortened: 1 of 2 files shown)


aaron97neu:
This looks great. What does `:latest` track? Master just like `:next` or the most recent `:<major>`?

aaron97neu:
> Pull request builds are not pushed to Dockerhub

I'm not sure about this. I like being able to track a PR while testing. IE I push an update to the PR, go to the test board, run `systemctl restart container-maintainer` and then it is running my new code. Instead of having to ssh into `oats2`, got into the `docker-compose.yml`, change the tag to the hash, ssh back into the board, run `systemctl restart ...`

abalmos:
> This looks great. What does `:latest` track? Master just like `:next` or the most recent `:<major>`?

In this setup `:latest` tracks the most recent release, no matter what type major/minor/patch.

> > Pull request builds are not pushed to Dockerhub
> 
> I'm not sure about this. I like being able to track a PR while testing. IE I push an update to the PR, go to the test board, run `systemctl restart container-maintainer` and then it is running my new code. Instead of having to ssh into `oats2`, got into the `docker-compose.yml`, change the tag to the hash, ssh back into the board, run `systemctl restart ...`

You would track `:feature-branch-name` now. I think that reads a bit better in the compose file too. I disabled pull request pushing because it will also fail when someone other than us opens a PR. Github strips all secrets (dockerhub token) from an action when it is triggered by an external user.

.github/workflows/main.yml
```diff
@@ -47,64 +48,79 @@ jobs:
...
+          echo 'Processing git ref:' $GITHUB_REF
+
           DOCKER_IMAGE=isoblue/${{ matrix.service }}
 
-          if [[ $GITHUB_REF == refs/tags/* ]]; then
-            VERSION=${GITHUB_REF#refs/tags/}
-          
+          # Release version is just the release number
+          if [[ $GITHUB_REF == refs/heads/release/* ]]; then
+            VERSION=${GITHUB_REF#refs/heads/release/}
```
aaron97neu, line 57: One future improvement could be to check for illegal chars here (slashes) or otherwise enforce the semantic versioning style. IE if it does not match `^[0-9]\.[0-9]\.[0-9]$` exit with an error

aaron97neu:
> You would track `:feature-branch-name` now.

Works for me

Also we might want to consider integrating this https://github.blog/2020-09-30-code-scanning-is-now-available/ in another PR

aaron97neu (approved):
Everything else looks good. We prob should make some issues for the improvements I mentioned 

:shipit:

facastiblancor (approved):

Issue #87, opened 2020-10-01
Add docker health checks


aaron97neu:
As part of our writing out tests, we should integrate health checks to our Dockerfiles
https://docs.docker.com/engine/reference/builder/#healthcheck

An example of a good health check for gps2tsdb would be a python script that gets the most recent gps point in the postgres db and ensures that it is no more than 5s old

aaron97neu:
Once #103 is merged, we will have gps, cell and can loggers with health checks. This leaves `can_watchdog`, `oada_upload`, and `socketcand`. Unsure if we can or want to do `socketcand` since it is developed by someone else.

@abalmos @facastiblancor what are your thoughts on how to check `can_watchdog` and `oada_upload` on a conceptual level? 
`can_watchdog` does not modify anything so measuring it's effects would be hard
`oada_upload` would also be interesting. You could check to see if the sent column changed recently but then you would be setting`oada_upload` as unhealthy if it does not have internet and isn't uploading stuff even if it is otherwise working fine. Maybe combine that with an internet check? Issue then is that it could be broken but you would never know if you are not connected to the internet. Is `oada_upload` conceptually unhealthy if it does not have internet?

facastiblancor:
@aaron97neu I think `socketcand` does not need a health check. It will be eventually phased out by other messaging systems in the future. But if we wanted to do a health check, it would be to check the program is running and the desired CAN interfaces are exposed. In my opinion, it would be better to implement this check in the apps using `socketcand`. For `can_watchdog` I think we don't need to check its health status. If it fails, then it will keep restarting. That way we know something is wrong with it.

PR #99, opened 2020-10-02
GPS2TSDB Health Check

aaron97neu:
Start of working on #87. This script checks the database and if it does not find a gps point that is less than 5 seconds old it fails

Using the default values for health checks (wait 30s to start checking, check every 30s, wait for 3 consecutive fails to change to unhealthy)

Should be trivial to put this on all the loggers. A health check for something like `oada_upload` will be another animal though

Examples:
You can check the monitor via `docker inspect`:
```
avena@avena-apalis-dev05:~$ docker inspect gps2tsdb_gps2tsdb_1 | head -n 50
[...]
            "Health": {
                "Status": "healthy",
                "FailingStreak": 0,
                "Log": [
                    {
                        "Start": "2020-10-02T22:16:22.978499033Z",
                        "End": "2020-10-02T22:16:24.287617231Z",
                        "ExitCode": 0,
                        "Output": "Database was last updated 0.994602 seconds ago, everything is normal ( threshold: 5 s )\n"
                    },
                    {
                        "Start": "2020-10-02T22:16:54.310686141Z",
                        "End": "2020-10-02T22:16:55.654777704Z",
                        "ExitCode": 0,
                        "Output": "Database was last updated 0.357015 seconds ago, everything is normal ( threshold: 5 s )\n"
                    },
     [..]
```
Health status in `docker ps`
```
avena@avena-apalis-dev05:~$ docker ps
CONTAINER ID        IMAGE                               COMMAND                  CREATED              STATUS                        PORTS                      NAMES
847d17f8d52c        gps2tsdb_gps2tsdb                   "python ./gps.py"        About a minute ago   Up About a minute (healthy)   0.0.0.0:10001->10001/tcp   gps2tsdb_gps2tsdb_1
```

Hopefully this will help our automated test scripts check the if the containers are performing normally

### DIFF
--- a/services/gps2tsdb/healthcheck.py
+++ b/services/gps2tsdb/healthcheck.py
@@ -1,0 +1,25 @@
+#!/usr/bin/python3
+import postgres
+import sys
+import os
+import datetime
+
+threshold = 5
+
+connectionurl='postgresql://' + os.environ['db_user'] + ':' + os.environ['db_password'] + '@postgres:' + os.environ['db_port'] + '/' + os.environ['db_database']
+db = postgres.Postgres(url=connectionurl)
+
+rst = db.one("SELECT * FROM gps ORDER BY time DESC LIMIT 1;")
+
+if rst == None:
+    print("Database has the wrong schema or no data");
+    sys.exit(1)
+
+lastupdate = rst[0]
+tdelta = datetime.datetime.now(lastupdate.tzinfo) - lastupdate
+sdelta = tdelta.total_seconds()
+if sdelta > threshold:
+  print('Database was last updated', sdelta, 'seconds ago, something is wrong ( threshold:', threshold, 's )');
+  sys.exit(1)
+
+print('Database was last updated', sdelta, 'seconds ago, everything is normal ( threshold:', threshold, 's )');
--- a/services/gps2tsdb/Dockerfile
+++ b/services/gps2tsdb/Dockerfile
@@ -38,6 +38,10 @@ COPY --from=builder /opt/venv /opt/venv
 # Make sure we use the virtualenv
 ENV PATH="/opt/venv/bin:$PATH"
 
+# Copy health check script
+COPY healthcheck.py .
+HEALTHCHECK CMD ["python", "./healthcheck.py"]
+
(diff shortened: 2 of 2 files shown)


services/gps2tsdb/healthcheck.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/python3
+import postgres
+import sys
+import os
+import datetime
+
+threshold = 5
```
facastiblancor, line 7: This could be an environment variable.

services/gps2tsdb/healthcheck.py
```diff
@@ -0,0 +1,25 @@
...
+
+if rst == None:
+    print("Database has the wrong schema or no data");
+    sys.exit(1)
+
+lastupdate = rst[0]
+tdelta = datetime.datetime.now(lastupdate.tzinfo) - lastupdate
+sdelta = tdelta.total_seconds()
+if sdelta > threshold:
+  print('Database was last updated', sdelta, 'seconds ago, something is wrong ( threshold:', threshold, 's )');
```
facastiblancor, line 22: Would it be possible to make systemd restart gpsd at this point? Currently we are fixing this issue by restarting gpsd from systemd. If we automated this, we could reduce downtime.

facastiblancor (approved):
So far, LGTM. I like it.

services/gps2tsdb/healthcheck.py
aaron97neu: We would have to change the container to give it access to the host's system systemd. Definitely possible, but I do not believe we have do it before and it breaks down another piece of the `sandbox` idea of containers

services/gps2tsdb/healthcheck.py
aaron97neu: Def could, however unsure if this needs to be something that is user configurable. I guess if you have a gps unit that gets points less than every 5 seconds?

services/gps2tsdb/healthcheck.py
abalmos: When we move `gpsd` to a container it could have a similar healthcheck which would restart itself

PR #103, opened 2020-10-09
Cell and can logger health checks

aaron97neu:
Additional health checks for a can and cell logger. Also moved all containers over to a `unless-stopped` restart policy as we discussed in a previous meeting

~Currently seeing a strange issue where the docker hub version does not seem to have been updated~ I'm dumb it works

### DIFF
--- a/services/.env
+++ b/services/.env
@@ -1,6 +1,0 @@
-gps_topic=gps
-tra_interface=REPLACE_WITH_TRA_INTERFACE_NAME
-isoblue_id=REPLACE_WITH_THINGSBOARD_DEVICE_ID
-mqtt_topic=v1/devices/me/telemetry
-mqtt_port=1883
-avena_domain=avena.ecn.purdue.edu
--- a/services/can_logger/Dockerfile
+++ b/services/can_logger/Dockerfile
@@ -34,6 +34,10 @@ COPY --from=builder /opt/venv /opt/venv
 # Make sure we use the virtualenv
 ENV PATH="/opt/venv/bin:$PATH"
 
+# Copy health check script
+COPY healthcheck.py .
+HEALTHCHECK CMD ["python", "./healthcheck.py"]
+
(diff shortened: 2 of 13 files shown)


aaron97neu:
Can't verify can-logger because it does not work in it's current state but the code is almost identical to cell-logger so it should be alright. Screenshot of everything else working:
image

facastiblancor:
> Can't verify can-logger because it does not work in it's current state but the code is almost identical to cell-logger so it should be alright. Screenshot of everything else working:
> image

That version of can-logger has issues with DB inserts, so that's probably the reason it is crashing. That issue has been solved and will be merged in an upcoming PR.

aaron97neu:
Awaiting @facastiblancor's changes to the cell-logger so I can ensure my health checks are compatible with his changes to the cell-logger

aaron97neu:
So merged master into this branch for testing and found that `can-logger` is unhealthy if there are messages appearing on one bus and no messages appearing on the other even if the logger itself is working fine. 

I am tempted to say that this situation is a configuration error and the unhealthy status is correct, but I could also watch the `rx_bytes` file and only check if that is incrementing. That would add some complexity as this should be stateless and I am not sure at the moment how to cheat that. Also at some point I am rewriting all of the logic in `can-logger` which kind of defeats the purpose of a simple health check

@facastiblancor thoughts?

abalmos:
The implement bus may go long periods of time with nothing on it and then someone attach an implement with a controller.

aaron97neu:
Good point. I'll plan on fixing this next week

aaron97neu:
Tests look good on my end, this should be ready for review

facastiblancor (approved):
Looks good to me. Cell check is OK. Personally, I prefer the sleep approach on the CAN logger, as it has less footprint.

abalmos (approved):
Algo LGTM

PR #104, opened 2020-10-14
CAN interface configuration in Ansible

### LINKED ISSUE #94: Test CAN interface configuration in Ansible.
This issue is to track the pending testing on a previous PR. Testing consists on deploying an ISOBlue unit with the new configuration tasks in Ansible.

### LINKED ISSUE #97: Validate CAN interfaces in can_logger
Implement checks on the available CAN interfaces to ensure valid interfaces are opened by the logger. Validations would be similar as the ones made by can_watchdog.

facastiblancor:
This PR fixes some issues with Ansible when CAN options are being configured. Closes #94  

TO DO:

- [x] Deal with systemd breaking ansible playbook. When Ansible is running any task using the "state: started" tag for systemd, the playbook will fail with "System state is unknown".


aaron97neu:
Further information on the systemd issue: https://github.com/ansible/ansible/issues/71528

It appears to be an issue with systemd in `245.6` that was fixed in `245.7`. `245.7` is not available in backports yet, but my bullseye/sid laptop is on `246` right now so a bit unclear about that. Last known working version appears to be `241-7~deb10u4`.

abalmos:
Looks like 246.6 is in backports for all architectures, expect `armhf` ...
 
https://packages.debian.org/search?keywords=systemd&searchon=names&section=all&suite=buster-backports

abalmos:
Wait, `armhf` has 254.6. Can't we just upgrade to that?

@facastiblancor @aaron97neu

aaron97neu:
245.7 is the version with the fix. 245.6 does not have the fix and the version we got when updating to backports

abalmos:
Yikes. Completely misread your comment.

abalmos:
Looks like systemd can only be built by gcc 9 or greater on arm due to the addition of this flag:

https://github.com/systemd/systemd/commit/1ad6056239b70ce69ea70108cf4e49a05a9c2a1d

Buster is on gcc 8, therefore no new versions of systemd can be built for buster backports.

It was suggested [here](https://developer.arm.com/documentation/dui0774/a/compiler-command-line-options/-mgeneral-regs-only) that -mfpu=none would be the same for armhf, but I don't really know. The commit that added the broken flag was addressing a specific bug in systemd-bootarm.efi

Looks like systemd build ci only tests against gcc 10

abalmos:
Started by opening an issue with systemd: https://github.com/systemd/systemd/pull/17934

abalmos:
@facastiblancor 247 is now in buster-backports

facastiblancor:
Great! Thank you @abalmos. Now we only need to test if that solves this longstanding issue.

ansible/avena/roles/can/tasks/can.yml
```diff
@@ -0,0 +1,48 @@
...
+        group: root
+        mode: "0644"
+      notify: reload systemd
+
+    - name: ensure can interfaces are started
+      become: yes
+      systemd:
+        name: can@{{ item }}
+        state: started
+      with_items: "{{ oats_can_interfaces }}"
```
aaron97neu, line 38: You are still using the old `oats_can_interfaces` variable  instead of `can_interfaces` here and on line two which keeps the configuration from working properly

aaron97neu (changes requested):
Runs without errors but need to fix the two varibles to test if configuring works for real

ansible/avena/roles/can/tasks/can.yml
facastiblancor: Thanks for pointing that out. I just fixed this and it worked in one ISOBlue I just deployed. More tests will have to be in order to confirm this works

facastiblancor:
Settings seem to work in a board I just deployed. After some more testing, this will be ready to merge.

aaron97neu (approved):
Approving pending further testing is good

facastiblancor:
Variable names changed again after misunderstanding. `oats-can-interfaces` is meant to store settings for an individual target, while `can-interfaces` is a group variable that stores the name of the CAN interfaces for all targets. Also, a task was added to restart the `can@` services after settings are changed. This has been tested after 5 deploys with different settings and no issues have been found.

aaron97neu (approved):
LGTM but lets move to configuring this with docker instead of ansible asap

PR #110, opened 2021-01-11
Upgrade OADA Helper Packages

aaron97neu:
Fixes new axios security issue https://github.com/axios/axios/pull/3410.
Updates to client should also fix this: https://github.com/OADA/client/issues/5

Update - Running new version on feeder at ault's farm. Full stack appears to be working fine. Will let it run through the end of the week to detect any intermittent issues and then will merge


aaron97neu:
Week of testing suspend cycles in a real environment plus successful testing with the newest versions of OADA should be sufficient. Merging

PR #111, opened 2021-01-21
OADA dependencies upgrade again

aaron97neu:
Upgrade to all OADA dependencies, moving away from beta releases in a few cases. A few other deps are out of date so going to try and upgrade those that do not cause problems


aaron97neu:
Testing on dev board looks good, awaiting results from board in the field in the next couple of days. Will merge if that comes back successful

Issue #122, opened 2021-02-16
Pin kernel updates until sufficently tested


aaron97neu:
On 2/15/20 I ran into an severe kernel panic issue after updating our development boards and restarting them. After about a day of troubleshooting with the rest of the team, we found that it was most likely due to [this issue](https://lists.infradead.org/pipermail/linux-arm-kernel/2021-February/635920.html) and rolling back the kernel from `5.10.x` to `5.9.x` successfully fixed it.

Our deployment procedure worked as intended. This issue was found in the lab and we froze `apt` related updates for field units until it is solved. However, this exercise pointed out several improvements we make in Avena and our procedures:

1. Automatic kernel rollbacks on failure to boot (Is this possible?)
1. Pinning kernel updates of in-field units until it is verified on a lab unit
   1.  [How to pin updates](https://www.snel.com/support/how-to-block-package-and-kernel-updates-in-debian/)
   1. Currently we do this, but it is pretty loosey goosey and runs on the honor system. ("Don't run the update script on the field boards until you have run it on the dev boards") Some sort of flag that is set on in-field units that checks to see if the new kernel has been tested before updating would be nice. Implementation details have not been considered
   1.  In order to do this, we need a pretty robust testing setup that _ensures_ problems that aren't immediately at boot are found before approving the update. As of writing this is in progress but far from satisfactory

PR #124, opened 2021-02-22
Ansible libseccomp2 and kernel locking

aaron97neu:
These fixes will allow the kernel version to be locked and installs the backports version of libsecomp2

Locking the kernel can come in handy for units in remote fields where risky upgrades like kernel upgrades are not desired. This addresses #122 at the most basic level, but could be done much more elegantly and with a few more features
The kernel version can be specified in the host ansible config file with the following line (or similar):
```
kernel_lock_version: '5.9.6-1~bpo10+1'
```

The libseccomp2 fix solves the issue found here: https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.13.0#time64_requirements

@facastiblancor and I are upgrading some of the lab units. If these are successful and all tests pass, we will merge this PR

### DIFF
--- a/ansible/avena/roles/core/tasks/update.yml
+++ b/ansible/avena/roles/core/tasks/update.yml
@@ -25,8 +25,8 @@
   become: yes
   apt:
     name:
-      - linux-image-armmp
-      - linux-headers-armmp
+      - linux-image-armmp{{ '='+kernel_lock_version if kernel_lock_version is defined }}
+      - linux-headers-armmp{{ '='+kernel_lock_version if kernel_lock_version is defined }}
     state: latest
     default_release: "{{ 'buster-backports' if ansible_distribution_release == 'buster' | default(omit) }}"
   notify: reboot system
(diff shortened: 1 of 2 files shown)


abalmos:
@aaron97neu Is it possible to somehow get the versions of everything installed and automatically lock everything? This would give us reproducible build, but make dev easy (lifting the locks, upgrading, testing, recomputing the locks).

aaron97neu:
@abalmos I don't think so. For example, running the following command to see all available kernels shows that only the most recent is available for install
```
$ apt-cache madison linux-headers-armmp
linux-headers-armmp | 5.10.13-1~bpo10+1 | http://deb.debian.org/debian buster-backports/main armhf Packages
linux-headers-armmp | 4.19+105+deb10u9 | http://http.us.debian.org/debian buster/main armhf Packages
linux-headers-armmp | 4.19+105+deb10u9 | http://security.debian.org/debian-security buster/updates/main armhf Packages
linux-latest | 105+deb10u9 | http://http.us.debian.org/debian buster/main Sources
linux-latest | 105+deb10u9 | http://security.debian.org/debian-security buster/updates/main Sources
```

This implies to me that a certain package version is not guaranteed to be available at any given time after it is released. The kernel could be a special case, but this does mean that I will need to rework the kernel aspect of this PR, especially because the `dist-upgrade` task immediately after will also attempt to upgrade the kernel

aaron97neu:
Doing one final test and will merge

PR #126, opened 2021-03-05
Add: dolly service

wang701:
Add a service that runs rclone periodically to upload data to box using cron.


services/dolly/docker-compose-example.yml
```diff
@@ -0,0 +1,19 @@
...
+    environment:
+      type: 'box'
+      box_sub_type: 'enterprise'
+      token: 'example-token'
+      override: '0'
+      bwlimit: '500k'
+      numcheckers: '2'
+      clvl: '0'
+    volumes:
+      - /data/log:/data:rw
```
abalmos, line 15: Should be a volume mount? Or at least show that as an example as we should likely be doing it that way anyway.

abalmos: Can we reduce to just `:r`?

services/dolly/go-dolly
```diff
@@ -0,0 +1,17 @@
+#!/bin/ash
+
+timestamp=`date +%Y/%m/%d-%H:%M:%S`
+echo "go-dolly: I was called at $timestamp"
+
+config_path="/config/rclone/rclone.conf"
+data_path="/data/test"
```
abalmos, line 7: I think this should just be `/data` ?

Maybe `/sync` or something like that would be nicer?

services/dolly/go-dolly
```diff
@@ -0,0 +1,17 @@
...
+remote_name="remotebox"
+remote_path="isoblue_rclone"
```
abalmos, line 9: This should be setable in docker-compose

services/dolly/go-dolly
```diff
@@ -0,0 +1,17 @@
...
+
+if [ ! -f $config_path ] || [ "$override" == "1" ]
+then
+  mkdir -p /config/rclone
+  echo -en "[$remote_name]\ntype = ${type}\nbox_sub_type = ${box_sub_type}\ntoken = ${token}" > $config_path
+fi
+
+rclone copy --progress --local-no-check-updated --compress-level=$clvl --checkers=$numcheckers --bwlimit=$bwlimit --log-level DEBUG $data_path $remote_name:$remote_path
```
abalmos, line 17: What if `$clvl`, `$bwlimit`, `$numcheckers` 

Do we always want --log-level DEBUG in production?

services/dolly/go-dolly
aaron97neu: Little concerned about doing just `/data` for isoblues that already have 10+gb files... might just have to be vigilante to delete those first

services/dolly/go-dolly
aaron97neu: Is there an clean in bash to make these values default but overwritable with envrion vars

services/dolly/docker-compose-example.yml
wang701: I thought we are already mounting `/data/log` to `/data` within the container? Not too familiar with docker so I can be totally wrong.

services/dolly/go-dolly
wang701: Agree with @aaron97neu . Doing `/data/test` just for the demoing purposes. But yes, as @abalmos suggested, `/sync` will be a better idea.

services/dolly/go-dolly
wang701: `$clvl` is compression level. Set to 0 to disable compression.
`$bwlimit` is bandwidth limit in bytes/second.
`$numcheckers` is the number of checkers running in parallel for verifying the files on the server and the backend side. More the better but I set it to 2 because I suspect it was pegging the CPU.

services/dolly/docker-compose-example.yml
abalmos: The test deploy maps a volume to `/data` but we could also map a local FS path to /data. Either way, does dolly need `rw` access to /data to work? Seems like dolly should only  be reading from that path.

services/dolly/go-dolly
abalmos: `/data` is the path in the container not the host.

abalmos: I think dolly should just sync whatever is at the container's `/data` path. The deploy should put the correct files there. For example, I suspect we may have a compose file like

```yml
services:
   dolly:
      image: isoblue/dolly
      volumes:
         - /path/to/some/data/on/local/fs:/data/mydata
         - can_data:/data/can
         - cell_data:/data/cell
```

services/dolly/go-dolly
abalmos: https://stackoverflow.com/a/2013589

or if live like @awlayton 

https://stackoverflow.com/a/28085062

I would argue that:

- `config_path` should be fixed and internal ... unless we will provide a way to completely override the auto generation of the  config
- `data_path` should be fixed and internal (your job is to put data there, see earlier comment on volume mounting) 
- `remote_name` is entirely internal I believe (?)
- `remote_path` is required input and dolly should fail if not provided (dolly should not just put data in your cloud storage where it wants)

services/dolly/go-dolly
abalmos: Sorry I wanted to say, "what if `$clvl`, `$bwlimit`, `$numcheckers` were not set". I think `rclone` likely dies with a help print out?

Seems like these are okay to have reasonable defaults for.

services/dolly/go-dolly
aaron97neu: Got confused with the very similar container and host paths... My bad. I agree with @abalmos this is the way to go. Maybe replacing `data` in the container with `/sync` to ease the confusion

aaron97neu:
Forgot to mention - @wang701 please add your ownership of this container to the end of the `CODEOWNERS` file [here](https://github.com/oats-center/isoblue-avena/blob/feature/dolly/.github/CODEOWNERS#L49) in the form `services/dolly     @wang701`

This means that you will be automatically added to PR's that modify code in this contianer

PR #127, opened 2021-03-05
Add: CAN CSV log rotation

facastiblancor:
This PR adds hourly CSV log rotation to `can_logger`. Logs will be stored in `/data/log/can` to avoid filling the data root with CAN logs. Also, it will create this folder on the first run if it was not previously created.

TO DO:
- [x] Test on actual ISOBlue Hardware.

### DIFF
--- a/services/can_logger/can_logger.py
+++ b/services/can_logger/can_logger.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-
 import socket
 import postgres
 import csv
@@ -8,51 +7,64 @@ import time
 import sys
 import collections
 import multiprocessing as mp
+import logging
 from datetime import datetime
 from psycopg2 import OperationalError
 from psycopg2.extras import execute_values
 
-
 def write_to_csv(wr_buff, can_bus):
-
     # Write to CSV from write buffer. First item is timestamp, second is
-    # can_id and third is can_data
-
-    with open(f'/data/log/{can_bus}.csv', mode='a') as logfd:
-
-        logcsv = csv.writer(logfd, delimiter=',', quotechar='"',
-                            quoting=csv.QUOTE_MINIMAL)
-
-        logcsv.writerows(wr_buff)
+    # can_bus, third is can_id, and last is can_data
+    # Generate timestamp to rotate the log
+    global log_path
+    timestamp = rotate_log()
+    while(not log_path):
+        try:
+            with open(f'/data/log/can/{can_bus}-{timestamp}.csv', mode='a') as logfd:
+                logcsv = csv.writer(logfd, delimiter=',', quotechar='"',
+                                    quoting=csv.QUOTE_MINIMAL)
+                logcsv.writerows(wr_buff)
+                log_path = True
+        except(FileNotFoundError):
+            path = '/data/log/can'
+            os.mkdir(path)
+            print(f'Logging path set to {path}')
 
     print(f'{can_bus}: Wrote successfully to CSV')
 
+def rotate_log():
+    global firstRun
+    t = time.strftime('%H',time.localtime())
+    # Check if enough time has passed to rotate the log
+    if(firstRun):
+        timestamp = time.strftime('%m-%d-%Y-%H-%M',time.localtime())
+        last_time = t
+        firstRun = False
+    elif(t == last_time):
+        pass
+    else:
+        last_time = t
+        # Save new timestamp
+        timestamp = time.strftime('%m-%d-%Y-%H-%M',time.localtime())
+
+    return timestamp
 
 def write_to_db(db, wr_buff, can_bus):
-    
     try:
-
         with db.get_cursor() as cursor:
-            
             execute_values(cursor,
                            "INSERT INTO can(time, can_interface, can_id, can_data) \
                            VALUES %s;", wr_buff)
-
     except(Exception, SyntaxError):
-
         print(f'{can_bus}: An error occured while inserting data to database')
-
     else:
-
         print(f'{can_bus}: Wrote successfully to DB')
 
 
 def db_init():
-
     connection_url = ('postgresql://' + os.environ['db_user'] + ':' +
                       os.environ['db_password'] + '@postgres:' +
                       os.environ['db_port'] + '/' + os.environ['db_database'])
-
     print('Initializing Postgres Object...')
     db = postgres.Postgres(url=connection_url)
     print('Ensuring timescaledb ext. is enabled')
@@ -67,7 +79,7 @@ def db_init():
 
     print("Ensuring can data table is a timescaledb hypertable")
     db.run("""
-           SELECT create_hypertable('can', 'time', if_not_exists => TRUE,  
+           SELECT create_hypertable('can', 'time', if_not_exists => TRUE,
            migrate_data => TRUE);""")
 
     print("Finished setting up tables")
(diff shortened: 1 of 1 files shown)


facastiblancor:
Added "syslog" style rotation and rotation interval settings via environment variables in `docker-compose.yml `. Log rotation intervals should be configured using `frequency` (integer) and `rotation interval` (Uppercase letter, S for seconds, M for minutes, H for Hours and D for days) env. variables. Needs more testing and adding user input checks.

services/can_logger/can_logger.py
```diff
@@ -210,6 +254,8 @@ def log_can(can_interface):
 host_port = os.environ['socketcand_port']
 host_interfaces = os.environ['can_interface']
 log_to = os.environ['log']
+frequency = int(os.environ['frequency'])
+interval = os.environ['rotation_interval']
```
aaron97neu, line 258: So if we wanted to rotate logs every 4 hours, `interval` would be `'h'` and `frequency` would be `4`? or would that mean we rotate 4 times an hour?  We should clarify these environmental variable names

services/can_logger/can_logger.py
```diff
@@ -32,22 +32,65 @@ def write_to_csv(wr_buff, can_bus):
...
+        dt = 86400
+        interval_str = 'day(s)'
     else:
-        last_time = t
-        # Save new timestamp
-        timestamp = time.strftime('%m-%d-%Y-%H-%M',time.localtime())
+        # Set interval to blank if the specified interval is not valid
+        interval = ''
+    # Enter to the loop if the interval is not blank
+    while(interval):
```
aaron97neu, line 58: Assuming that this is equivalent to `while interval != '':`, I would prefer the explicit comparison for clarity. As it is, it implies to me that interval is a boolean

services/can_logger/can_logger.py
```diff
@@ -32,22 +32,65 @@ def write_to_csv(wr_buff, can_bus):
...
-    elif(t == last_time):
-        pass
+def rotate_log(can_bus, interval, frequency):
+    last_time = 0
+    f = 0
+    firstRun = True
+    # Check for a valid interval. If valid, set dt (update interval) to its
+    # equivalent in seconds, i.e., 60 seconds for minute rotation and 3600s for
+    # hourly rotation.
+    if(interval == 'S'):
```
aaron97neu, line 42:
```suggestion
    if(interval.lower() == 's'):
```
Wouldn't mind making the units case insensitive

services/can_logger/can_logger.py
```diff
@@ -32,22 +32,65 @@ def write_to_csv(wr_buff, can_bus):
...
-    # Check if enough time has passed to rotate the log
-    if(firstRun):
-        timestamp = time.strftime('%m-%d-%Y-%H-%M',time.localtime())
-        last_time = t
-        firstRun = False
-    elif(t == last_time):
-        pass
+def rotate_log(can_bus, interval, frequency):
+    last_time = 0
+    f = 0
```
aaron97neu, line 37: Wouldn't mind a more readable variable name here or a comment explaining

aaron97neu (commented):
Few requested changes for readability, logic appears to be sound. Appreciate the extensive comments

services/can_logger/can_logger.py
facastiblancor: Good point. Actually it means we are rotating every 4 hours. For this case, `period` sounds much better.

facastiblancor:
Addressed all comments. Pending for testing on ISOBlue hardware. If no further problems are identified, this should be ready to merge.

services/can_logger/can_logger.py
```diff
@@ -8,51 +7,124 @@
...
-                            quoting=csv.QUOTE_MINIMAL)
-
-        logcsv.writerows(wr_buff)
+    # can_bus, third is can_id, and last is can_data
+    # Generate timestamp to rotate the log
+    global log_path
+    #timestamp = rotate_log()
+    while(not log_path):
+        try:
+            with open(f'/data/log/can/{can_bus}.csv', mode='a') as logfd:
```
abalmos, line 23: This should either not be hard coded or a simple path. Maybe both?

default /data, overridden by `DATA_DIR=<path>` or `--data-dir=<path>` or something

The reason is the service should not know anything about the host system. It is the only thing in the container and so it "owns" the data folder. If the deployment wants files logged to the host filesystem, then they will map their path of choice to /data in the container or (more correctly due to user id / permissions issues with local FS mapping) just map /data to a volume.

services/can_logger/can_logger.py
```diff
@@ -8,51 +7,124 @@
...
-    # can_id and third is can_data
-
-    with open(f'/data/log/{can_bus}.csv', mode='a') as logfd:
-
-        logcsv = csv.writer(logfd, delimiter=',', quotechar='"',
-                            quoting=csv.QUOTE_MINIMAL)
-
-        logcsv.writerows(wr_buff)
+    # can_bus, third is can_id, and last is can_data
+    # Generate timestamp to rotate the log
```
abalmos, line 18: Not done anymore?

services/can_logger/can_logger.py
```diff
@@ -8,51 +7,124 @@
...
+    global log_path
+    #timestamp = rotate_log()
+    while(not log_path):
```
abalmos, line 21: Can't this complexity be removed by checking if the log path exists, and if not, creating it before opening the file?

services/can_logger/can_logger.py
```diff
@@ -8,51 +7,124 @@
 import sys
 import collections
 import multiprocessing as mp
+import logging
 from datetime import datetime
 from psycopg2 import OperationalError
 from psycopg2.extras import execute_values
 
-
 def write_to_csv(wr_buff, can_bus):
```
abalmos, line 15: We open the log file for every message written?

services/can_logger/can_logger.py
```diff
@@ -8,51 +7,124 @@
...
+                logcsv = csv.writer(logfd, delimiter=',', quotechar='"',
+                                    quoting=csv.QUOTE_MINIMAL)
+                logcsv.writerows(wr_buff)
+                log_path = True
+        except(FileNotFoundError):
+            path = '/data/log/can'
+            os.mkdir(path)
+            print(f'Logging path set to {path}')
 
     print(f'{can_bus}: Wrote successfully to CSV')
```
abalmos, line 33: This should be debug, I assume this creates a lot of stdout that ends up written to the disk by docker.

services/can_logger/can_logger.py
```diff
@@ -8,51 +7,124 @@
...
 
+def rotate_log(can_bus, interval, period):
```
abalmos, line 35: Somewhat of a nit: wouldn't accepting `interval` as seconds directly remove complexity and increase capability?

services/can_logger/can_logger.py
```diff
@@ -154,163 +205,128 @@ def log_can(can_interface):
...
             db_started = False
-
         else:
-
             db_started = True
 
 for can_bus in can_interfaces:
     print('Creating process for', can_bus)
     mp.Process(target=log_can, args=(can_bus,)).start()
+    mp.Process(target=rotate_log, args=(can_bus, interval, period,)).start()
```
abalmos, line 332: I think this can be really simplified if you just roll the log based on the can messages timestamp -- this might even be more ideal because then CAN messages are in the correct file based on time.

For every RX'd message, compare the time to the next roll over time. If it is larger, close the current file, open a new file with the current roll over time, compute the next roll over time, and continue the processing loop. You avoid the race condition games the current design is playing with `os.rename` while still logging to the old file handle and you don't need anything `mp.Process` either.

abalmos (commented):
I think my review comes from a perspective of a different design, so take it all for what it is worth. 

I would argue we need to reconsider this design, however, if it is working well enough right now, then we should just move on because it will be replaced in the NATS version of things.

services/can_logger/can_logger.py
facastiblancor: Indeed. However, this is printed once at most. Even in the scenario that the container is continually failing, the path will be created the first time the program runs and handles the exception. After that, the program would not raise that exception anymore.

services/can_logger/can_logger.py
facastiblancor: No. We open every time the write_buffer is full (which right now stores up to 1000 messages). So depending on the message rate, the file will be more of less frequently opened.

services/can_logger/can_logger.py
facastiblancor: It is true. However, the way I thought the log rotation depends on `time.strftime(time.time())`, which depending on the formatting options (set by the flag -INTERVAL) will return the numeric value of the current seconds, minutes and, days. Then we use this output to compare against a value previously stored in order to keep track of interval changes. This is likely to change in the future for the NATS-based logger.

services/can_logger/can_logger.py
facastiblancor: Right. That is a leftover comment of old code.

services/can_logger/can_logger.py
facastiblancor: Indeed. This can be further included for a future NATS-based version.

services/can_logger/can_logger.py
facastiblancor: Absolutely. However, to implement rotation based on messages it is necessary to change the codebase of the current logger, as we are not logging by single messages, but instead, we are logging batches of thousands of messages.  This is also part of the logic that needs improvement and will be improved for a NATS-based logger.

services/can_logger/can_logger.py
facastiblancor: Yes. Ideally, right after starting the program.

facastiblancor:
Testing on ISOBlue hardware has shown good results. However, migrating from socketcand to NATS requires a complete rewrite of this program. Therefore, this code will be deprecated as soon as a NATS-based logger is released. Comments and suggestions will be addressed in a NATS-based logger. This is ready to be merged if no further discussion is needed.

services/can_logger/can_logger.py
abalmos: I think this is referring to line 33

facastiblancor: My bad. I though it referred to the green highlighted code. Indeed, this will be removed.

abalmos: Good point. I don't understand why they are green?

aaron97neu (approved):
Agree that this will be deprecated with the NATS version. As long as it works I am fine with it. Hopefully we can introduce some simplifications to the logic in the NATS version, this is starting to turn into a hairy mess

PR #131, opened 2021-03-18
WIP SMS Messaging 

aaron97neu:
Works as is, I would consider this the MVP. 

Still would like to to enable the ability to text the service to "subscribe" to alerts. All twilio examples use flask which I would like to avoid, currently thinking I am going to try to use the `requests` library with async but unsure of the feasibility of that


facastiblancor:
Merging to master in order to do a code clean up. Code will be archived in the 'legacy' folder

PR #137, opened 2021-04-16
remove oats_can_interfaces variable

aaron97neu:
Should fix sleep cycling issue

### DIFF
--- a/ansible/avena/roles/can/templates/system-sleep/reset-can-interfaces.j2
+++ b/ansible/avena/roles/can/templates/system-sleep/reset-can-interfaces.j2
@@ -5,8 +5,7 @@ case $1/$2 in
         post/*)
                 echo "Waking up from $2..."
                 {% for interface in can_interfaces %}
-                systemctl restart can@{{ interface }}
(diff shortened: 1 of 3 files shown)


ansible/avena/roles/can/templates/system-sleep/reset-can-interfaces.j2
```diff
@@ -4,7 +4,7 @@ echo "Reset can interfaces"
 case $1/$2 in
         post/*)
                 echo "Waking up from $2..."
-                {% for interface in oats_can_interfaces %}
+                {% for interface in can_interfaces %}
                 systemctl restart can@{{ interface }}
```
abalmos, line 8:
```suggestion
                systemctl restart can@{{ interface.name }}
```

abalmos (approved):
Looks right less that one loop

facastiblancor (approved):
Looks good. Let's wait for some testing before merging.

aaron97neu:
Tested on two boards in the lab, should be good to merge

PR #146, opened 2021-06-23
Add linux/arm64

abalmos:
Add ARM 64 to container builds to support Pi 4

@Torin1914 please comment after testing.


facastiblancor:
Looks good to me. Will approve and merge after @Torin1914 tests the images.

facastiblancor (approved):
@Torin1914 Tested and was able to pull the images with the new tag. We need to agree on a strategy to update the images of the containers that are still on PRs.

PR #167, opened 2021-11-24
[WIP] Add: Avena sdrd service

facastiblancor:
This PR adds sdrd, a docker-container with the Ettus Research UHD and GNURadio libraries to support running GNURadio-generated scripts inside a container.

TO DO:

- [ ] Test service with USB USRPs (B200mini)
- [ ] Test with network-based USRPs (N210, X310, USRP1&2, etc.)
- [ ] Write a NATS code for Avena-compliant status checking.

### DIFF
--- a/.github/workflows/build-and-push-services.yml
+++ b/.github/workflows/build-and-push-services.yml
@@ -9,17 +9,17 @@ jobs:
     strategy:
       matrix:
         service:
-          - gps2tsdb
-          - gps_nats
-          - oada_upload
-          - cell_logger
-          - can_watchdog
-          - can_logger
-          - socketcand
-          - cand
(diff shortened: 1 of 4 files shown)


abalmos:
Is the goal of this container to be used as a base layer for other containers?.

facastiblancor:
Kind of. For the moment, this is more a wrapper of both UHD and GNURadio able to run scripts exported from block diagrams. It could be made into a base container for future services that require access to SDRs, but would need to rethink a bit the way it works.

abalmos:
How do you add the scripts?

facastiblancor:
Will close this PR, as most of the work has been superseeded by the Avena FM service

PR #168, opened 2021-12-03
Bugfix/oada upload/fix sleep

aultac:
This should fix the sleep issue that causes CPU to peg w/ oada_upload.


abalmos:
I'm not sure what the CI will do with this branch name. If it breaks, you may need to name it like [this](https://github.com/oats-center/isoblue-avena/blob/master/CONTRIBUTING.md#feature-branches-and-pull-requests).

abalmos:
@aultac Looks like the image is ready for testing with `image: isoblue/oada_upload:bugfix-oada_upload-fix_sleep`

aultac:
Fabio told me what the branch name needed to be, and was quite adamant that it had to have this format specifically (bugifx/service_name/bugfixname) or the CI would not work.  I don’t know which of you is right :).

Aaron

abalmos:
@aultac The format is as described in the contributing document. `bug-name` can be anything, but I was concerned somewhere I spit on `/` and that having a second would break things. It appears to have worked though.

.github/workflows/build-and-push-services.yml
```diff
@@ -16,8 +16,8 @@ jobs:
           - can_watchdog
           - can_logger
           - socketcand
-          - cand
-          - j1939d
+          # - cand
+          # - j1939d
```
abalmos, line 20: Why?

services/oada_upload/dev-docker-compose.yml
```diff
@@ -0,0 +1,20 @@
+version: '3.1'
```
abalmos, line 1: What is the point of this file?

services/oada_upload/dev-startup.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/bash
+
+echo "-----------------------------------------------------"
+echo "build: "
+yarn run build
```
abalmos, line 5: Wny building in a container at runtime? Should be built in the Dockerfile if a build stage is needed.

services/oada_upload/package.json
```diff
@@ -1,24 +1,30 @@
...
   "name": "@isoblue/oada_upload",
+  "type": "module",
   "version": "0.0.1",
   "description": "Upload Avena time-series datapoints to an OADA store",
-  "main": "src/oada_upload.ts",
+  "exports": "./build/oada_upload.js",
   "repository": "https://github.com/oats-group/isoblue-avena",
   "author": "Aaron Neustedter",
   "contributors": [
     "Andrew Balmos <abalmos@purdue.edu>"
```
abalmos, line 10: Should add yourself

services/oada_upload/package.json
```diff
@@ -1,24 +1,30 @@
...
   ],
   "license": "MIT",
   "scripts": {
-    "start": "ts-node src/oada_upload.ts"
+    "oldstart": "ts-node --files src/oada_upload.ts",
```
abalmos, line 14: oldstart?

services/oada_upload/src/atomic-sleep.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'atomic-sleep';
```
abalmos, line 1: I think a proper type would be something like this. I haven't pulled, but does the code pass typechecks?

```suggestion
declare module "atomic-sleep" {
  function sleep(t: number): void;

  export default sleep;
}
```

services/oada_upload/src/oada_upload.ts
```diff
@@ -15,15 +15,19 @@ declare global {
...
 import ksuid from 'ksuid';
 import { randomBytes } from 'crypto'
 import pMap from 'p-map';
 import { connect } from '@oada/client';
 import { V1 as Location } from '@oada/types/oada/isoblue/location/v1';
+import sleep from 'atomic-sleep';
+import GeohashLib from 'latlon-geohash';
 
-import { isoblueDataTree } from './trees';
+import { isoblueDataTree } from './trees.js';
```
abalmos, line 28: .js ?

services/oada_upload/tsconfig.json
```diff
@@ -14,6 +14,6 @@
     "forceConsistentCasingInFileNames": true,
     "allowSyntheticDefaultImports": true
   },
-  "include": ["src/**/*"],
+  "imclude": [ "src/**/*" ],
```
abalmos, line 17: typo? I'm not sure the `include` line is even though -- I believe the default is fine.

abalmos (approved):
I don't know this code, but a quick glance though seems okay. I left notes a few things.

This service is not in use anywhere, so I think it is safe to merge when @aultac wants. @aultac can clean up any leftover bugs after live testing. I suggest at least a quick sanity check of the published image on a dev ISOBlue.

.github/workflows/build-and-push-services.yml
facastiblancor: Rust emulated compile takes forever and times out (giving failure messages). We can uncomment them as soon as the code is ready to merge.

aultac:
@fabiocastiblancor is handling this, the PR was only because he thought it needed a PR in order for the CI to build it.   It shouldn’t be merged in until he has tested it this week.  He’s been giving me and Prof. Evans a tutorial on how to code a service for Avena, and we started w/ attempting to fixing the bug in oada_upload that caused the CPU to peg.

In response to your question about why those two services were commented out, Fabio says that have to be commented out or the CI times out so we commented them out.

Aaron

.github/workflows/build-and-push-services.yml
abalmos: I see. I just figured we could ignore errors for services not in this PR. Need to get the new CI pipeine here now that you can share workflows in GHA.

services/oada_upload/dev-docker-compose.yml
aultac: I wish the author of the service had made this in the first place.  This file is what you'd copy to "docker-compose.yml" to dev on this service locally.  docker-compose.yml is .gitignored so that you can't accidently commit your local dev copy of it to the repo, but this is basically the "example" docker-compose.yml for a developer.

services/oada_upload/dev-startup.sh
aultac: Good point, I was just trying to navigate TS nonsense and finally got things working by building and then running rather than using ts-node which is what was originally there.  ts-node has an issue with declaration files that just doesn't work at all.  It would make sense now that things are working to move the build step to the Dockerfile rather than the entrypoint, although when developing and mapping your local code into the container, it is better to just build in the entrypoint so you can just restart the service on code changes.

services/oada_upload/package.json
aultac: that's the command that was used to start the service when we started working on it, but that command quit working so it got saved there.  It can be safely deleted now, good catch.

services/oada_upload/src/oada_upload.ts
aultac: ts-node refused to import it if you left off the .js.  I'm not sure if that's also still required for building with tsc instead of ts-node, but it was related to the mixture of commonjs and esm between the imported libraries and the code.  Maybe this can be removed now.

services/oada_upload/tsconfig.json
aultac: you are right, it is a typo

services/oada_upload/dev-docker-compose.yml
aaron97neu: > I wish the author of the service had made this in the first place.

Right here: https://github.com/oats-center/isoblue-avena/blob/master/services/oada_upload/docker-compose-example.yml

aultac: @aaron97neu this dev-docker-composel.yml is used when developing locally on your machine rather than on an isoblue.  It maps the local filesystem in to /code so you can just edit and restart to see the container run, and doesn't depend on isoblue-specific other services like postgres.  i.e. I can just run a container on my machine and it will connect to a postgress on a running isoblue over wiregaurd before deploying to the isoblue.  The original example docker-compose is what you'd use when adding it to a docker-compose deployment.

What I meant by "I wish the author had made this in the first place" wasn't "man, @aaron97neu really dropped the ball :)" , I meant that we should make it standard procedure for isoblue services to have a dev docker-compose.yml file so you can easily just grab a service and start it on your machine straight away to dev on it when debugging.  You did a good job.

PR #189, opened 2022-04-13
Python files for ISO Blue from Capstone project

ksoontho:
Changed Dockerfile, Removed gps-nats, changed requirements, changed folder name.


abalmos:
This deletes `gps_nats` folder. Could you bring it and the code in it back and push again ?

services/CapstonePython/PythonFinal.py
```diff
@@ -0,0 +1,112 @@
+import asyncio
+import nats
+import json
+import time
+
+
+async def main():
+    # Connect to NATS!
+    nc = await nats.connect("172.16.254.5")  # change ip to connect to correct isoblue
```
abalmos, line 9: This needs to be `localhost` not an ip

abalmos (changes requested):
Need to bring back gps_nats and address my review comment and then we'll run the CI to see if the docker image builds

ksoontho:
> This deletes `gps_nats` folder. Could you bring it and the code in it back and push again ?

Will fix that tomorrow morning.

PR #208, opened 2022-06-12
Avena FM demo

facastiblancor:
This is a GNURadio block diagram that implements a FM radio tuner controlled by NATS.


services/avena_fm/Dockerfile
```diff
@@ -0,0 +1,19 @@
+# Use Fedora as base distribution
+FROM fedora:36
+MAINTAINER 'OATS Center'
+
+# Install updates
+RUN dnf -y update; dnf -y update; dnf -y clean all
```
abalmos, line 6: Update twice?

services/avena_fm/requirements.txt
```diff
@@ -0,0 +1 @@
+pynats2
```
abalmos, line 1: Does this need a locked version to avoid random breakage later?

services/avena_fm/scripts/avena_fm_demo_epy_block_0.py
```diff
@@ -0,0 +1,46 @@
...
+            name='NATS Sink',   # will show up in GRC
+            in_sig=[(np.ushort, vector_size)],
+            out_sig=None
+        )
+
+        #hostname = os.getenv('HOSTNAME')
+        self.vector_size = vector_size
+        self.freq = freq
+        self.span = span
+        self.subject = os.getenv('HOSTNAME') + '.' + subject
```
abalmos, line 32: AVENA_PREFIX maybe?

services/avena_fm/scripts/avena_fm_demo_epy_block_0.py
```diff
@@ -0,0 +1,46 @@
...
+        self.nc = NATSClient(nats_server, socket_timeout=2)
+        self.nc.connect()
+        
+    def work(self, input_items, output_items):
+        
+        b64encpayload = str(b64encode(input_items[0][0]), 'utf-8')
+        json_dump = json.dumps({'fft': b64encpayload,
+                                'fc' : self.freq, 
+                                'span' : self.span,
```
abalmos, line 41: I think I'll need gain to compute the FFt y-axis and set the slider properly

services/avena_fm/scripts/avena_fm_demo_epy_block_1.py
```diff
@@ -0,0 +1,60 @@
...
+            in_sig=[np.complex64],
+            out_sig=None
+        )
+        # if an attribute with the same name as a parameter is found,
+        # a callback is registered (properties work, too).
+        self.port_1 = 'Frequency'
+        self.port_2 = 'Center Frequency'
+        self.port_3 = 'Gain'
+        self.port_4 = 'Stream'
+        self.subject = os.getenv('HOSTNAME') + '.' + subject
```
abalmos, line 25: AVENA_PREFIX?

services/avena_fm/scripts/avena_fm_demo_epy_block_0.py
abalmos: I suppose I'll need stream too.

That or we need a req/reply for me to ask for the current value of everything.

Seems easier for now to just send it all every time

abalmos (commented):
LGTM short a few small additions.

Might be useful to put a screen shot of the flow graph in git?

services/avena_fm/scripts/avena_fm_demo_epy_block_0.py
facastiblancor: Both audio and FFT streams stop when stream is False.

abalmos: I see. 

This has some problems still too, like if I close the app with it on then it will just stream and use data anyway.

services/avena_fm/scripts/avena_fm_demo_epy_block_1.py
facastiblancor: So, should we set AVENA_PREFIX to be equal to HOSTNAME by default?

PR #209, opened 2022-06-12
Add: Hostname prefix on gps_nats messages

facastiblancor:
This PR adds a prefix with the machine hostname to all messages published by this service.


services/gps_nats/gps.py
```diff
@@ -32,15 +31,13 @@ async def run(loop):
...
-    #await nc.connect("nats://nats:4222")
+    nc = await nats.connect("localhost")
 
     for new_data in gps_socket:
         if new_data:
             fix = json.loads(new_data)
-            subject = "gps." + str(fix["class"])
-            print("Publishing new data point to subject", subject, ": ", new_data[:-1])
+            subject = os.getenv('HOSTNAME') + ".gps." + str(fix["class"])
+            #print("Publishing new data point to subject", subject, ": ", new_data[:-1])
```
abalmos, line 40: Maybe PREFIX or AVENA_PREFIX. In the multiple user/application case we would need more than one per host.

Small, but the first string join could be factored out of the hot loop.

facastiblancor:
> Does the python nats lib expect you to flush?

Yes, that is to ensure that the message has been published.

abalmos:
> Yes, that is to ensure that the message has been published

Sure, it is just an antipattern to call it in the `nats.ws` and `nats-rs` library outside of writing tests, so I surprised to see it here. For example: https://github.com/nats-io/nats.ws#flush

PR #216, opened 2022-09-26
Change Prometheus prefix to be j1939 

facastiblancor:
This PR changes the Prometheus prefix to be 'j1939', instead of being customizable by the NATS_SUB environment variable. This is due to Prometheus being not compatible with NATS subject semantics, especially with the topic separation characters (aka dots)

### DIFF
--- a/.github/workflows/build-and-push-services.yml
+++ b/.github/workflows/build-and-push-services.yml
@@ -9,17 +9,17 @@ jobs:
     strategy:
       matrix:
         service:
-          - j1939_filter
-          - gps2tsdb
-          - gps_nats
-          - oada_upload
-          - cell_logger
-          - can_watchdog
-          - can_logger
-          - socketcand
-          - cand
+          #- j1939_filter
+          #- gps2tsdb
+          #- gps_nats
+          #- oada_upload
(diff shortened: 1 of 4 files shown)


abalmos (approved):
Unfortunate.

abalmos:
We should probably just use `cross` rather than all this custom vendoring. I had promised to fix this build chain, but never got to it.

Is this blocking you?

facastiblancor:
Wanted to try `cross` but couldn't see how to integrate it to our GH actions workflow. It is not blocking me right now. Just wanted to try adding cross-compilation to see if I could reduce build times for rust services. I might add a couple of services in the future and wanted to avoid the slow emulator-based build. If this does not work, I will revert to the original QEMU builder and merge the code that needed attention.

abalmos:
This is clo

facastiblancor:
Added a **very crude** cross-compiling routine for the j1939d service in the dockerfile. So far it works, but a proper GHA cross-compiling flow for Rust containers must be considered in the future.

facastiblancor:
Services now build correctly without errors, and the fix is working on our test stand units.

PR #262, opened 2023-07-20
Bug: Add exception handling to JSON parser

facastiblancor:
This PR adds exception handling to gps-nats, as some JSON strings seem to crash the Python JSON parser.

TODO:
- Test with RTK messages


facastiblancor:
This PR replaced the `gps3` client with `gpsdclient`, which performs better and is currently maintained. After testing with RTK, we now get all the messages generated by GPSD.

